PR #1, opened 2018-12-17
Feat/initialization

alaibe:
Embark 4 template with:
- React
- React Router
- Redux + Saga
- 1 Example feature
- Bootstrap 4
- Reactstrap


app/js/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { ConnectedRouter } from "connected-react-router";
+import { Provider } from 'react-redux';
+
+import 'bootstrap/dist/css/bootstrap.css';
```
bgits: This is just a personal preference, I stay away from global CSS as it becomes harder to manage conflicts

alaibe: @iurimatias suggested me to use bootstrap. hence why it is there with Reactstrap.

I would like to try https://www.styled-components.com.
Would that align more with your preference?
That would mean more control and a long term win but also not as easy to start.

@iurimatias what do you think?

bgits: styled-components are definitely aligned with this thinking.

PR #5, opened 2018-12-17
Query display price

bgits:


app/js/features/prices/reducer.js
```diff
@@ -0,0 +1,18 @@
...
+    };
+  default:
+    return state;
+  }
+}
+
+export default reducer;
+
+export const getEthUsdPrice = state => state.prices.ETH.USD
+export const getSntUsdPrice = state => state.prices.SNT.USD
```
alaibe: I usually use selector for that, this is personal preference, feel free to dismiss

package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "bootstrap": "^4.1.3",
     "connected-react-router": "^6.0.0",
+    "cryptocompare": "^0.7.0",
```
iurimatias: we should use etherscan instead which seems to be what status is currently using right now.

PR #9, opened 2018-12-18
Add google map with fake data

jrainville:
image


package.json
```diff
@@ -20,9 +20,11 @@
     "embark-solium": "0.0.1",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
+    "react-google-maps": "^9.4.5",
     "react-redux": "^6.0.0",
     "react-router-dom": "^4.3.1",
     "reactstrap": "^6.5.0",
+    "recompose": "^0.30.0",
```
alaibe: This lib can do a lot and also change the way we do components, that can be fun to include it

app/js/layout/App.jsx
```diff
@@ -2,11 +2,13 @@ import React, { Component, Fragment } from 'react';
 import { HashRouter, Route } from "react-router-dom";
 import { connect } from 'react-redux';
 import { Container } from 'reactstrap';
+import EmbarkJS from 'Embark/EmbarkJS'; // Needed for auto reload
```
alaibe: I trust you, could you just describe why?

jrainville: Here or in the comment in the code?

alaibe: I can see we are using EmbarkJS below so I get the import bu why the `// Needed for auto reload`

jrainville: Oh, because at the time of me coding, we weren't actually using EmbarkJS. I just rebased so I didn't see we are using it.
The reason why we need it for reload is because we add the Websocket snippet inside EmbarkJS and I haven't figured out how we can extract that snippet and put it inside the actual code.

alaibe: Understood do you think we should remove the comment?

jrainville: Yeah good idea

PR #10, opened 2018-12-19
feat: embarkjs as a service

alaibe:


app/js/layout/App.jsx
```diff
@@ -20,20 +22,12 @@ const relevantPairs = {
 class App extends Component {
   constructor(props) {
     super(props);
+    this.props.init();
```
bgits: It's not clear why `init` and `fetchPrices` need to called in the App component constructor. They don't rely on the App Component, or initialize local state or a binding.

alaibe: They don't have to be here and they could be wherever we want in the init process, AFAIK the difference will be negligible. 
I think that App is a good choice as most of the interaction with the redux store will be in Container.
So for principle of least surprise, putting it here makes sense.

However, if you strongly disagree, I can put those 2 call wherever you prefer

bgits: This is not a blocking issue it's just a personal distaste for bloated classes and separating views from models.

PR #14, opened 2018-12-20
feature(@teller-network/contracts) support to rate a transaction

iurimatias:


contracts/escrow.sol
```diff
@@ -162,4 +161,24 @@ contract Escrow is Pausable {
...
+     * @notice Requires contract to not be paused.
+     *         Can only be executed by the buyer
+     *         Transaction must released
+     */
+    function rate_transaction(uint _escrowId, uint _rate) public whenNotPaused {
+        require(_rate >= 1, "Rating needs to be at least 1");
+        require(_rate <= 5, "Rating needs to be at less than or equal to 5");
+
+        EscrowTransaction storage trx = transactions[_escrowId];
+
```
richard-ramos: what do you think of adding a require(trx.rating == 0) so people can set their rating only once?

jrainville: I think you might want to change your rating, like on Amazon, you can change your mind. 
I agree however that a transaction cannot break or something like an object can.

test/escrow_spec.js
```diff
@@ -328,4 +327,84 @@ contract("Escrow", function() {
...
+      if (isPaused) {
+        receipt = await Escrow.methods.unpause().send({from: accounts[0]});
+      }
+      receipt = await Escrow.methods.create(accounts[1], expirationTime).send({from: accounts[0], value});
+      created = receipt.events.Created;
+      escrowId = created.returnValues.escrowId;
+      await Escrow.methods.release(escrowId).send({from: accounts[0]});
+    })
+
+    it("should not allow a score last less than 1", async() => {
```
jrainville:
```suggestion
    it("should not allow a score that's less than 1", async() => {
```

test/escrow_spec.js
```diff
@@ -328,4 +327,84 @@ contract("Escrow", function() {
...
+      try {
+        receipt = await Escrow.methods.rate_transaction(escrowId, 0).send({from: accounts[1]});
```
jrainville: Do you really need to override the one above? You don't really reuse the receipts so you could keep them local.

test/escrow_spec.js
```diff
@@ -328,4 +327,84 @@ contract("Escrow", function() {
...
+      try {
+        receipt = await Escrow.methods.rate_transaction(escrowId, 6).send({from: accounts[1]});
+        assert.fail('should have reverted: should not allow a score to be more than 5');
+      } catch(error) {
+        TestUtils.assertJump(error);
+        assert.ok(error.message.indexOf('Rating needs to be at less than or equal to 5'))
+      }
+    })
+
+    for(let i=1; i<=5; i++) {
```
jrainville: I feel like doing 5 tests for that is a bit over kill. Just doing a test for 1 and 5 should be enough.

PR #19, opened 2018-12-22
feature(@teller-network) calculate and display user rating

iurimatias:


app/js/features/license/reducer.js
```diff
@@ -1,22 +1,28 @@
...
-      };
+      }};
     case CHECK_LICENSE_OWNER_SUCCEEDED:
-      return {
+      return {...state, ...{
         licenseOwner: action.isLicenseOwner
-      };
+      }};
+    case USER_RATING_SUCCEEDED:
+      return {...state, ...{
```
bgits: you could also do:
```
const { userRating } = action
return { ...state, userRating }
```

app/js/features/license/saga.js
```diff
@@ -34,4 +37,21 @@ export function *onCheckLicenseOwner() {
   yield takeEvery(CHECK_LICENSE_OWNER, doCheckLicenseOwner);
 }
 
-export default [fork(onBuyLicense), fork(onCheckLicenseOwner)];
+export async function *doUserRating() {
+  try {
+    const arrAvg = arr => arr.reduce((a,b) => a + b, 0) / arr.length;
+    const events = await Escrow.getPastEvents('Rating', {fromBlock: 1, filter: {seller: web3.eth.defaultAccount}});
```
bgits: The purpose of using sagas is to get the flexibility of generators as async/await is only one specific type of generator. Since this function is already a generator you can call here using sagas call and no need to make this an async function as well. 

Sagas explains it here: https://github.com/redux-saga/redux-saga/#redux-saga-chooses-generators-over-asyncawait

IMO, 90% of the time you don't actually need sagas so you could just define an async function to fetch then dispatch the rating or error.

PR #24, opened 2019-01-08
feature(@teller-network/contracts) adds arbitration to escrow

richard-ramos:


contracts/teller-network/Escrow.sol
```diff
@@ -24,25 +26,44 @@ contract Escrow is Pausable {
...
-    event Created(address indexed seller, address indexed buyer, uint escrowId);
+    address public arbitrer;
+
+    event Created(address indexed seller, address indexed buyer, uint escrowId, uint expirationTime);
     event Paid(uint escrowId);
+    event Released(uint escrowId);
     event Canceled(uint escrowId);
     event Rating(address indexed seller, address indexed buyer, uint escrowId, uint rating);
 
+    mapping(uint => ArbitrationCase) public arbitrationCases;
```
alaibe: Do you think Arbitration should be more like it's own contract
e.g Pausable, Arbitrable

contracts/teller-network/Escrow.sol
```diff
@@ -185,4 +278,119 @@ contract Escrow is Pausable {
         trx.rating  = _rate;
         emit Rating(trx.seller, trx.buyer, _escrowId, _rate);
     }
+
+    modifier onlyArbitrer {
```
alaibe: Is it normal there is no test for all the function below?

richard-ramos: I forgot to add tests for `setArbitrer()` and `isArbitrer()`.  Good catch!

contracts/teller-network/Escrow.sol
richard-ramos: I'll review to see how can this be refactored, since setArbitrationResult depends on Escrow specific function. I guess some kind of delegation should be possible.

PR #38, opened 2019-01-15
feat: add first unit test structure

alaibe:
Add test for:
* reducer
* actions
* component

As a first example I used the License

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["@babel/preset-env", "@babel/preset-react"],
+  "plugins": ["@babel/plugin-proposal-class-properties"]
+}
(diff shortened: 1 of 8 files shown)


alaibe:
That is why I used snapshot mainly, if you just change the text you can approve the change on the snapshot

I am not a fan to test ui too much too for that reason, snapshot are made for that I think

jrainville:
Ah ok, so you don't code the snapshot yourself? If that's so, it's good then!

alaibe:
Yes, if a snapshot changed while you run the test, you have an error,
but if you are confident with your changes, you can run:
```
yarn test:unit --updateSnapshot
```
And that will regenerate the snapshot

If you realize the changes are unexpected, then fix the component :)

PR #39, opened 2019-01-15
Add create escrow form validation

jrainville:
image

### DIFF
--- a/app/js/components/EscrowList.jsx
+++ b/app/js/components/EscrowList.jsx
@@ -5,7 +5,6 @@ import PropTypes from 'prop-types';
 import {getEscrowState, escrowStates} from "../features/escrow/helpers";
 import {SIGNATURE_PAYMENT, SIGNATURE_OPEN_CASE} from "../features/escrow/constants";
 import Rating from "./Rating";
-import moment from 'moment';
 import { withNamespaces } from 'react-i18next';
 import SignatureDialog from "./SignatureDialog";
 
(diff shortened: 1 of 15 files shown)


jrainville:
Rebased and added the library you suggested.
Only thing, I needed to keep the expiration manual validation because it is not an input but a custom component so the validator does not work on it

app/js/components/EscrowList.jsx
```diff
@@ -67,7 +68,7 @@ const EscrowList = (props) => (
             <td>{getEscrowStateText(escrow, props.t)}</td>
             <td>{escrow.buyer === web3.eth.defaultAccount ? escrow.seller : escrow.buyer}</td>
             <td>{escrow.amount}</td>
-            <td>{new Date(escrow.expirationTime * 1000).toString()}</td>
+            <td>{moment(escrow.expirationTime * 1000).toString()}</td>
```
alaibe: I can see we are doing this multiple times, do you think we should move this conversion to the selector?

app/js/features/escrow/reducer.js
```diff
@@ -40,6 +40,11 @@ const escrowBuilder = function (escrowObject) {
 
 function reducer(state = DEFAULT_STATE, action) {
   let escrows  = cloneDeep(state.escrows);
+  let currentEscrow;
+  if (action.escrowId) {
+    currentEscrow = escrows.find(escrow => escrow.escrowId === action.escrowId);
+  }
+
```
alaibe: What if currentEscrow is undefined at this stage (line 47)?
Is it even possible?

jrainville: It's possible that escrowId is available but it's a new escrow (eg creating a new escrow), but the reducer case doesn't use `currentEscrow` so it's not an issue there. 
We could add a condition on the cases that require `currentEscrow` and throw an error, but that's what's gonna happen anyway.

jrainville:
@alaibe Fixed and rebased

PR #40, opened 2019-01-17
Adds missing arbitration and address components to storybook + fixes properties displays

richard-ramos:


app/js/components/CreateEscrowForm.jsx
```diff
@@ -103,4 +103,5 @@ CreateEscrowForm.propTypes = {
   result: PropTypes.object
 };
 
+export const CreateEscrowFormComponent = CreateEscrowForm;
```
alaibe: I guess you are doing that for the PropTypes.
Do you think it would be better if doing:
```
export class CreateEscrowForm {
}

export default withNamespaces()(CreateEscrowForm);
```

Then when you require, you have 2 form:
```
import CreateEscrowForm from './CreateEscrowForm';
```
and
```
import {CreateEscrowForm} from './CreateEscrowForm';
```
you could also 

```
import {CreateEscrowForm as CreateEscrowFormWithProptypes} from './CreateEscrowForm';
```

Just a suggestion, not sure it is really better

stories/address.stories.jsx
```diff
@@ -0,0 +1,21 @@
...
+import {withInfo} from "@storybook/addon-info";
+
+import Address from '../app/js/components/Address';
+
+
+storiesOf('Address', module)
+  .add(
+    "Normal",
+    withInfo({inline: true})(() => (
+      <Address address="0x1122334455667788990011223344556677889900" compact={false} />
```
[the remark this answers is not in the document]
richard-ramos: Good idea!

app/js/components/CreateEscrowForm.jsx
richard-ramos: I tried this option and couldn't make it work with Storybook. If you have some free time, please show me an example with a single component and I'll replicate the same behavior with the others

alaibe:
@richard-ramos @jrainville I have updated the PR, let me know what you think

PR #44, opened 2019-01-22
feat: add home page

alaibe:
image


app/js/containers/HomeContainer.jsx
```diff
@@ -1,13 +1,39 @@
...
+          <Col xs={12} className="my-auto">
+            <h1 className="text-center">{t('home.welcome')}</h1>
+          </Col>
+        </Row>
+        <Row className="mt-5">
+          <Col xs={6}>
+            <Button tag={Link} color="info" block to="/buy">{t('home.buy')}</Button>
+          </Col>
+          <Col xs={6}>
+            <Button tag={Link} color="info" block to="/sell">{t('home.sell')}</Button>
```
jrainville: Do these links bring somewhere right now?

alaibe: Does an empty page count as somewhere?

Other pages have to be built :)

jrainville: Haha, yeah that's what I was expecting. It's fine for now. We could have pointed it to myEscrows for sell and the map for buy but it's gonna change anyway so no big deal

PR #45, opened 2019-01-22
Improve loading states by adding Loader and TX Hash

jrainville:
This will be particularly useful when we test on testnet and later on mainnet, because it gives better visual input that something is happening, with a spinner.
Also, when the TX is submitted and being mined, we show the TX hash and a link to etherscan.

image

image


app/js/features/escrow/saga.js
```diff
@@ -132,8 +197,20 @@ export function *rateTx({escrowId, rating}) {
   try {
     const toSend = Escrow.methods.rateTransaction(escrowId, rating);
     const estimatedGas = yield call(toSend.estimateGas);
-    yield call(toSend.send, {gasLimit: estimatedGas + 1000, from: web3.eth.defaultAccount});
-    yield put({type: RATE_TRANSACTION_SUCCEEDED, escrowId, rating});
+    const promiseEvent = toSend.send({gasLimit: estimatedGas + 1000, from: web3.eth.defaultAccount});
+    const channel = eventChannel(promiseEventEmitter.bind(null, promiseEvent));
+    while (true) {
```
alaibe: Do you think we could we wrap that into a function?
It seems we repeat always the same pattern when we send a transaction

app/js/features/signature/constants.js
```diff
@@ -1,4 +1,5 @@
 export const INCLUDE_SIGNATURE = 'INCLUDE_SIGNATURE';
+export const INCLUDE_SIGNATURE_PRE_SUCCESS = 'INCLUDE_SIGNATURE_PRE_SUCCESS';
 export const INCLUDE_SIGNATURE_SUCCEEDED = 'INCLUDE_SIGNATURE_SUCCEEDED';
```
alaibe: instead of PRE_SUCCESS and SUCCEEDED, maybe we should name them:
INCLUDE_SIGNATURE_TRANSACTION and
INCLUDE_SIGNATURE_RECEIPT

app/css/Form.scss
```diff
@@ -4,3 +4,12 @@
 .is-invalid-input {
   border-color: $danger;
 }
+
+@keyframes spinner {
```
alaibe: Do you think we could create a: TransactionForm component, that handle all the cases about loading/error/success
That way this scss would not be local but on the component itself?

app/css/Form.scss
```diff
@@ -4,3 +4,12 @@
...
+  to {transform: rotate(360deg);}
```
alaibe: I think font awesome has class that include spinning: `fa-spin`
(if I remember correctly)

app/css/Form.scss
```diff
@@ -4,3 +4,12 @@
...
+}
+
+.loading{
```
alaibe: Font awesome has classes for that AFAIK

app/js/components/ArbitrationList.jsx
```diff
@@ -18,7 +22,8 @@ const ArbitrationList = (props) => (
       <CardTitle>Disputed escrows</CardTitle>
     </CardHeader>
     <CardBody>
-      {props.loading && <p>Loading...</p>}
+      {props.loading && <p><FontAwesomeIcon icon={faSpinner} className="loading"/>{props.t('arbitrationList.loading')}</p>}
+      {props.txHash && <TransactionHash txHash={props.txHash}/>}
```
alaibe: Do you think we could have a component using render props or HOC (pattern up to you) 
that abstract all of that?

app/js/features/escrow/saga.js
jrainville: Sadly no, because the `yields` need to be in the same scope as the original function. It's really the big problem with the current way sagas work, you can't call actions from within another function and we need them because it uses events.

app/css/Form.scss
jrainville: Ok yeah, I'll try

app/css/Form.scss
jrainville: Yeah I remember seeing that indeed.

app/js/features/escrow/saga.js
alaibe: ohhh :cry: I see what you mean, this is also quite annoying.
Do you think we could make them generic enough to have like:
https://github.com/embark-framework/embark/blob/master/embark-ui/src/sagas/index.js#L11-L19

For all transaction call?

PR #47, opened 2019-01-27
feat: currencies selection and margin rate

richard-ramos:
Allows sellers to select which currencies to sell and what margin rate above and below market to use.
To access these screens: http://localhost:8000/#/seller/fiat 

Items pending / for discussion
1. Wizard: this implementation is naive, performing flow step changes by pushing addresses to history
2. There's no way to identify which currencies are FIAT in cryptocompare API
3. To display the rate between FIAT and crypto, the "Assets" screen is required.
4. Is only crypto to FIAT the only valid operation for sellers?

Components will be added to storybook in separate PR


app/js/components/FiatSelectorForm.jsx
```diff
@@ -0,0 +1,103 @@
+import React, {Component, Fragment} from 'react';
+import PropTypes from 'prop-types';
+import {FormGroup, FormFeedback} from 'reactstrap';
+import {Typeahead} from 'react-bootstrap-typeahead';
+import '../../../node_modules/react-datetime/css/react-datetime.css';
+import '../../css/Form.scss';
```
alaibe: may be we should import those 2 files globally? in `index.js`?

app/js/components/MarginSelectorForm.jsx
```diff
@@ -0,0 +1,124 @@
...
+    this.setState({isAbove: e.target.value === "above"});
+  }
+
+  onInputChange = (e) => {
+    this.setState({rate: e.target.value});
+  }
+
+  componentDidMount() {
+    // Redirecting to previous step
+    if(!this.props.fiat) this.props.history.push('/seller/fiat');
```
alaibe: we will need a real wizard here in future

app/js/components/MarginSelectorForm.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+    this.setState({
+      isAbove: this.props.margin.isAbove,
```
alaibe: Cannot be done in constructor?

app/js/components/MarginSelectorForm.jsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  _submit(address){
+    if(!this.validate()) return;    
+    this.props.onSubmit(parseInt(this.state.rate, 10), this.state.isAbove); 
+    this.props.history.push(address);
+  }
+
+  back = () => {
+    this._submit("/seller/fiat");
```
alaibe: Should you validate and submit on back?

richard-ramos: It can be removed. I added it to store the values on the form if you click back, but now that you mention it, it does not look like normal UX. I'll remove this

app/js/components/MarginSelectorForm.jsx
richard-ramos: Yup! :1st_place_medal:

app/js/components/MarginSelectorForm.jsx
richard-ramos: Indeed. I'll refactor this using the wizard component from the other PR

app/js/components/FiatSelectorForm.jsx
```diff
@@ -0,0 +1,103 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      fiat: '',
+      error: ''
+    };
+  }
+
+  componentDidMount(){
+    if(this.props.value) this.setState({ fiat: this.props.value });
```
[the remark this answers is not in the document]
richard-ramos: yup! :+1:

app/js/components/FiatSelectorForm.jsx
richard-ramos: Good idea!

app/js/components/MarginSelectorForm.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+    return ( 
+      <Form ref={c => { this.form = c; }}>
+        {error && <Alert color="danger">{error}</Alert>}
+        <FormGroup>
+            <Label for="rate">{t('marginSelectorForm.margin')}</Label>
+            <InputGroup>
+              <Input type="text" name="rate" id="rate" placeholder="0" className="form-control"
+                  value={rate} onChange={this.onInputChange} validations={[required, isInteger]} />
+              <InputGroupAddon addonType="append" style={{height: "calc(2.25rem + 2px)"}}>%</InputGroupAddon>
```
jrainville: Why do you need to specify the height manually? I thought it should use the same classes so be the same height by default.

richard-ramos: When a validation error for that field is shown, the InputGroupAddon extended vertically to the size of the input field + error message. I force the height to be a static value to avoid this.

jrainville: ooooh, ok thanks for the clarification. Maybe we could put that in `Form.scss` so that we don't have that problem with any other inputs

PR #48, opened 2019-01-28
feat: add wizard base component

alaibe:


alaibe:
> Wizard path diverges depending on decisions taken in a step (i.e. If you selected if you selected value X it will go straight to step 9, or if you selected value Y, the wizard path is different

I'd say that if path diverges, then the last step is another wizard?
A wizard can have a nested wizard

> You load a wizard step directly by writing the path on the browser and you're not supposed to reach this page if you haven't set the values from previous steps.

I am sure we could develop a jump function that will go to the desired step. Also I feel we should be careful with this

Maybe using nesting wizard would be better

PR #49, opened 2019-01-30
Add Seller workflow and profile

jrainville:
Seller profile made by Anthony.

I did the seller workflow. Currently, it saves on page change to the redux state.

Doesn't have the "Submit" at the end, because we need to adapt the Wizard for that and we don't know yet what's after that.


app/js/containers/Seller/SellerContactContainer.jsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  changeNickname = (nickname) => {
+    this.validate(this.state.contactCode, nickname);
+    this.setState({nickname});
+  };
+
+  render() {
+    return (
+      <Fragment>
```
alaibe: You don't need fragment if you return only 1 component.
Also we should probably try to find what are the components that are reusable SellerContact instead of doing a 1-1 with the container
(we can probably refactor that later)

app/js/containers/Seller/SellerPaymentMethodContainer.jsx
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+      selectedMethods.push(selectedMethod);
+    }
+    this.setState({selectedMethods});
+    this.validate(selectedMethods);
+  };
+
+  render() {
+    return (
+      <Fragment>
```
alaibe: same as before for fragment

app/js/hoc/withFooter.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+    disableNext = () => {
+      this.setState({ nextEnabled: false });
+    };
+
+    change() {
+      this.changeSubs.forEach(cb => {
+        cb.call(cb);
+      });
+      this.changeSubs = [];
```
alaibe: What does this one do?

app/js/hoc/withFooter.scss
```diff
@@ -0,0 +1,14 @@
+$footer-height: 58px;
+
+.container {
+  padding-bottom: $footer-height;
```
alaibe: It is a bit scary to change the container class, as we may use it in lot of places.
Do you think we could have a wrapper dedicated class?

app/js/index.scss
```diff
@@ -9,4 +9,8 @@ body {
 
 body > div { 
   flex: 1; 
+}
+
+.container {
+  overflow: auto;
```
alaibe: is it still needed?

app/js/hoc/withFooter.jsx
jrainville: This is a very simple event fake bus. It's nice to only save the state to the redux state only on page change, so we register a callback with `onPageChange` and then, when the footer changes the page, it calls this.
We clear the array of callbacks after, because they should only be called once.

alaibe: Nice 👍

alaibe: Oh yes of course, I said **may be later**. I was just dropping idea here

app/js/hoc/withFooter.scss
jrainville: SO, I just changed the footer a bit so that it stays at the bottom withotu being fixed. It took way too long for what it does, but at least it works now. 

I still need the padding, because otherwise, the content goes under the footer.
However, it's not super scary, because it's bottom and also, it's just a padding, so if there is nothing under the container (like always I think), it will never be seen.

PR #50, opened 2019-02-01
Add buyer workflow components (map branch)

jrainville:
Adds the components and screens for the buyer workflow using the Map. So we still need the bank transfer/list workflow, but it is similar.

This doesn't save to the Redux state, because I'm still unsure of what we actually have to save and what ar ethe wanted values. For example, do we have an offer ID? What's the value to identify the seller? License ID or address?

Anyway, it makes this PR lighter to be only UI.

image
image
image
image


stories/map.stories.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withInfo } from "@storybook/addon-info";
 
-import Map, { Map as MapWrapped } from '../app/js/components/Map';
```
alaibe: The same way we have stories in seller directory, should you also move buyer into his own directory

PR #55, opened 2019-02-05
Feat/seller store

alaibe:


contracts/teller-network/SellerStore.sol
```diff
@@ -0,0 +1,145 @@
...
+        address _asset,
+        address _statusContractCode,
+        string memory _location,
+        string memory _currency,
+        string memory _username,
+        PaymenMethods[] memory _paymentMethods,
+        MarketType _marketType,
+        uint8 _margin
+    ) public {
+        require(License(license).isLicenseOwner(msg.sender), "Not a license owner");
```
jrainville: Need to check if license is still valid too, I think.

alaibe: If the license is released, then this method will returns false

alaibe:
@jrainville good point I will then separate seller and offers

contracts/teller-network/SellerStore.sol
```diff
@@ -39,19 +39,27 @@ contract SellerStore is Ownable {
...
 
     address public license;
     Seller[] public sellers;
-    mapping(address => mapping (uint256 => bool)) public ownerToSellers;
+    Offer[] public offers;
+    mapping(address => bool) public sellerWhitelist;
+    mapping(address => uint256) public addressToSeller;
+
+    mapping(address => mapping (uint256 => bool)) public offerWhitelist;
+    mapping(address => uint256[]) public addressToOffers;
```
jrainville: Not sure what the whitelists are for? If we have a mapping with the address pointing to the offerID, can't we just check for the existence there and not have a whitelist? I probably just don,t know something about solidity here.

alaibe: I'll try to explain what I know and @richard-ramos can correct me if I am wrong:

if you have a mapping: `mapping(address => uint256) foo`
and access `foo[msg.sender]` if there is nothing in the mapping it will return the default value which will be 0.
However 0 is a valid ID.
So having a white list (mapping with boolean) I can be sure when it is true or false by setting it explicitly to true and the default value being false

Makes sense?

jrainville: Yeah makes sense. Can't we just start IDs at 1 then?

alaibe: IDs are not controlled by me but are indexes of array

I think that would add a lot of overhead:
* Need to add an empty seller/offer in array by default to avoid id 0
* Rewrite public getter to ignore this id 0
...
And probably others.
I have seen this technique quite a lot and seems to be the suggested way of doing this kind of manipulation.
However I might be wrong what do you think @richard-ramos

jrainville: Ah ok. Will wait on to see what Richard says, but what you're saying makes a lot of sense

richard-ramos: If to be a seller, we require a license, isn't it possible to remove the need for a sellerWhitelist by calling `isLicenseOwner` from the license contract?

alaibe: You can be a license owner but not be a seller yet

alaibe: how would you do that?

contracts/teller-network/SellerStore.sol
```diff
@@ -0,0 +1,173 @@
...
+        require(License(license).isLicenseOwner(msg.sender), "Not a license owner");
+        require(_margin <= 100, "Margin too high");
+
+        if (!sellerWhitelist[msg.sender]) {
+            Seller memory seller = Seller(_statusContractCode, _location, _username);
+            uint256 sellerId = sellers.push(seller) - 1;
+            addressToSeller[msg.sender] = sellerId;
+            sellerWhitelist[msg.sender] = true;
+        } else {
+            sellers[addressToSeller[msg.sender]].statusContractCode = _statusContractCode;
```
richard-ramos: if we assign sellers[addressToSeller[msg.sender]] to a temporary variable like  `Seller storage tmpSeller`, and then refer to that tmp variable, does the cost increases / decreases?

contracts/teller-network/SellerStore.sol
richard-ramos: I'm a little confused due to this code here:
```
if (!sellerWhitelist[msg.sender]) {
            Seller memory seller = Seller(_statusContractCode, _location, _username);
            uint256 sellerId = sellers.push(seller) - 1;
            addressToSeller[msg.sender] = sellerId;
            sellerWhitelist[msg.sender] = true;
        }
```
According to this: ` addressToSeller[msg.sender] `  will always contain the latest `sellerId` right? Is it possible that I as a seller can update different seller records except the latest, because in `update` we are always updating the latest seller data record.

richard-ramos: Or another question:
can a Seller have more than one  
```
struct Seller {
        address statusContractCode;
        string location;
        string username;
    }
``` 
?

jrainville: Don't think so, that's 1 to 1.

alaibe: It is a 1-1 so one address can only have 1 seller this is why I do:
```
if (!sellerWhitelist[msg.sender]) {
  Seller memory seller = Seller(_statusContractCode, _location, _username);
  uint256 sellerId = sellers.push(seller) - 1;
  addressToSeller[msg.sender] = sellerId;
  sellerWhitelist[msg.sender] = true;
}
```

richard-ramos: I think in this case, we can remove the need for an array and withelist like this.   i.e.: 

    
```
mapping(address => Seller) public sellers;


function add(
        address _asset,
        address _statusContractCode,
        string memory _location,
        string memory _currency,
        string memory _username,
        PaymenMethods[] memory _paymentMethods,
        MarketType _marketType,
        uint8 _margin
    ) public {
        require(License(license).isLicenseOwner(msg.sender), "Not a license owner");
        require(_margin <= 100, "Margin too high");

        
        sellers[msg.sender].statusContractCode = _statusContractCode;
        sellers[msg.sender].location = _location;
        sellers[msg.sender].username = _username;

        
        Offer memory offer = Offer(_asset, _currency, _margin, _paymentMethods, _marketType);
        uint256 offerId = offers.push(offer) - 1;
        offerWhitelist[msg.sender][offerId] = true;
        addressToOffers[msg.sender].push(offerId);

        emit Added(msg.sender, offerId, _asset, _statusContractCode, _location, _currency, _username, _paymentMethods, _marketType, _margin);
    }
```
what do you think?

contracts/teller-network/SellerStore.sol
alaibe: The cost decrease slightly

contracts/teller-network/SellerStore.sol
alaibe: I wanted to do that but then how do you access a list of seller?

richard-ramos: You're right!

The more I look at this, the greater the need I see to have this information stored offchain instead of using a contract. Filtering the data will be difficult using this model.

alaibe:
@jrainville @richard-ramos are you happy with current implementation?

alaibe:
The last commit is adding the status to the offer and the rates to the seller

contracts/teller-network/MetadataStore.sol
```diff
@@ -0,0 +1,212 @@
...
+        address _asset,
+        address _statusContractCode,
+        string memory _location,
+        string memory _currency,
+        string memory _username,
+        PaymenMethods[] memory _paymentMethods,
+        MarketType _marketType,
+        uint8 _margin
+    ) public {
+        require(sellerWhitelist[msg.sender], "Seller do not exists");
```
jrainville:
```suggestion
        require(sellerWhitelist[msg.sender], "Seller does not exist");
```

contracts/teller-network/MetadataStore.sol
```diff
@@ -0,0 +1,212 @@
...
+        require(offerWhitelist[msg.sender][_offerId], "Offer do not exists");
```
jrainville:
```suggestion
        require(offerWhitelist[msg.sender][_offerId], "Offer does not exist");
```

contracts/teller-network/MetadataStore.sol
```diff
@@ -0,0 +1,212 @@
...
+
+        emit Updated(msg.sender, _offerId, _asset, _statusContractCode, _location, _currency, _username, _paymentMethods, _marketType, _margin);
+    }
+
+    /**
+    * @dev Rate a seller from 0 to 4
+    * @param _to The address to rate
+    * @param _value The rate value
+    */
+    function rate(address _to, RateValue _value) public {
```
jrainville: We already have a rating for escrows. Do we transfer it here? It's probably easier to implement if it's inside the Seller, but it might make less sense, because you can rate someone you never had an escrow with.

contracts/teller-network/MetadataStore.sol
```diff
@@ -0,0 +1,212 @@
...
+        require(_to != msg.sender, "Cannot rate yourself");
```
jrainville: If we do have the rating here, we need to make sure you only rate once (or we handle changing a vote)

PR #56, opened 2019-02-05
feat(buyer-workflow): add buyer list workflow and improve components

jrainville:
Add fonts, fix offer listing, add redux save, and more


app/js/index.js
```diff
@@ -6,6 +6,7 @@ import { I18nextProvider } from 'react-i18next';
 import i18n from './i18n';
 
 import 'bootstrap/dist/css/bootstrap.css';
+import '../css/fonts/Inter/inter.css';
```
alaibe: if you want the storybook to have the fonts too, you need to add it into `.storybook/config.js` too

app/js/components/Buyer/SorterFilter.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
+import {faAngleDown, faSortAmountDown} from "@fortawesome/free-solid-svg-icons";
+
+const SorterFilter = () => (
+  <div className="sorterselect font-weight-bold px-3 py-2 bg-secondary rounded v-align-center my-3">
+            <span className="sort-icon bg-dark text-white rounded-circle d-inline-block text-center p-2 mr-2">
```
alaibe: Little indentation issue here

app/js/containers/Buyer/OfferListContainer.jsx
```diff
@@ -1,13 +1,53 @@
...
+  {assets: [{name: 'ETH', price: 123}, {name: 'SNT', price: 3}], seller: 'George', location: 'Berlin, Deutschland', nbTrades: 54, positiveRatings: 43, negativeRatings: 23},
+  {assets: [{name: 'ETH', price: 123}, {name: 'SNT', price: 3}], seller: 'Allison', location: 'Rio de Janeiro', nbTrades: 2, positiveRatings: 43, negativeRatings: 23},
+  {assets: [{name: 'ETH', price: 123}, {name: 'SNT', price: 3}], seller: 'Roger', location: 'Paris, France', nbTrades: 76, positiveRatings: 43, negativeRatings: 23}
+];
 
 class OfferListContainer extends Component {
   render() {
-    return (
-      <OfferList/>
+          return (
```
alaibe: once again, indentation issue

app/js/components/Buyer/OfferTrade.jsx
```diff
@@ -1,40 +1,58 @@
...
 import {Row, Col, FormGroup, Button} from 'reactstrap';
 import Form from 'react-validation/build/form';
 import Input from 'react-validation/build/input';
 import Blockies from 'react-blockies';
-import {isNumber} from "../../validators";
+import {isNumber, lowerEqThan, higherEqThan} from "../../validators";
 
-const OfferTrade = ({address, name, min, max, fiat, asset}) => (
+const OfferTrade = ({address, name, min, max, fiat, asset, onClick, assetQty, fiatQty, onFiatChange, onAssetChange, disabled, t}) => (
   <Row>
```
alaibe: I think this component can be split

app/js/containers/Seller/SellerMarginContainer.jsx
```diff
@@ -25,7 +25,6 @@ class SellerMarginContainer extends Component {
   }
 
   validate(newMargin) {
-    console.log(newMargin);
```
alaibe: :100:

app/js/features/buyer/reducer.js
```diff
@@ -0,0 +1,32 @@
+import {SET_CONTACT, SET_OFFER_ID, SET_TRADE} from './constants';
+
+const DEFAULT_STATE = {};
+
+function reducer(state = DEFAULT_STATE, action) {
```
alaibe: It is possible to write test for reducer too

app/js/features/buyer/actions.js
```diff
@@ -0,0 +1,7 @@
+import {SET_CONTACT, SET_OFFER_ID, SET_TRADE} from './constants';
```
alaibe: It is possible to write test for action too

alaibe: Even if for those one that might not be needed as they are really simple, up to you!

app/js/containers/Buyer/Listing.scss
```diff
@@ -0,0 +1,32 @@
+@import "../../../../node_modules/bootstrap/scss/functions";
+@import "../../../../node_modules/bootstrap/scss/variables";
+
+$logo-size: 64px;
+.logo-circle {
```
alaibe: If it is not wrapped around container name I feel this need to be extracted as a component

app/js/containers/Buyer/Listing.scss
```diff
@@ -0,0 +1,32 @@
...
+      border-bottom-right-radius: 15px;
+    }
+
+    span:first-child {
+      border-radius: 15px 0 0 15px;
+    }
+  }
+}
+
+.sort-icon {
```
alaibe: If it is not wrapped around container name I feel this need to be extracted as a component

app/js/features/buyer/reducer.js
```diff
@@ -0,0 +1,32 @@
...
+  switch (action.type) {
+    case SET_CONTACT:
+      return {
+        ...state, ...{
+          nickname: action.nickname,
```
alaibe: Is it really buyer? I would have imagine all of that is part of the profile?
I think we have some overlap between buy and sell and that should be only one.

app/js/components/Buyer/OfferTrade.jsx
jrainville: I think I could split the Form out, but it makes it so that we have a lot of props to pass and just makes the app heavier in terms of files. Do you think it's really worth it? Especially since I don't think we will reuse that component.

alaibe: The idea would be to pass less props, for example in this one I think I can see 2 components being called from the container:

```
<Col xs="12" className="mt-5 text-center">
      <h2>Trade amount with <br/><span><Blockies seed={address || '0x0'} className="rounded-circle"/> {name}</span></h2>
      <p className="mt-3">Min: {min}{fiat.symbol} - Max: {max}{fiat.symbol}</p>
    </Col>
```
and
```
<Col xs="12" className="mt-4">
      <Form className="text-center">
        <FormGroup>
          <Input type="number" name="fiat" className="form-control" value={fiatQty}
                 validations={[isNumber, lowerEqThan.bind(null, max), higherEqThan.bind(null, min)]}
                 placeholder="Fiat quantity" onChange={(e) => onFiatChange(e.target.value)}/>
          <span className="input-icon">{fiat.id}</span>
        </FormGroup>
        <FormGroup>
          <Input type="number" name="asset" className="form-control" value={assetQty} validations={[isNumber]}
                 placeholder="Asset quantity" onChange={(e) => onAssetChange(e.target.value)}/>
          <span className="input-icon">{asset}</span>
        </FormGroup>
        {disabled && <p className="text-info">{t('buyer.offerTrade.enterBefore')}</p>}
        <Button color="primary" className="px-4" onClick={onClick} disabled={disabled}>Open trade</Button>
      </Form>
    </Col>
```

If you think it is not worth it, it is fine, up to you!

app/js/features/buyer/reducer.js
jrainville: I'll wait until refactoring those, because I'm still unsure how we handle it. In the end, we might even just use the Status nickname, so no need for the form.

Depends also if we do save the name and of the contact code in the contract. If yes, having them inside buyer simplifies the gathering of the information.

Anyway, we'll see.

PR #58, opened 2019-02-08
Add filter menu UI

jrainville:
This is just UI components. It does nothing

image

### DIFF
--- a/app/js/components/Buyer/SorterFilter.jsx
+++ b/app/js/components/Buyer/SorterFilter.jsx
@@ -1,17 +1,90 @@
-import React from 'react';
+import React, {Component, Fragment} from 'react';
 import PropTypes from 'prop-types';
 import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
 import {faAngleDown, faSortAmountDown} from "@fortawesome/free-solid-svg-icons";
+import classnames from 'classnames';
 
 import './SorterFilter.scss';
+import CheckButton from '../ui/CheckButton';
+import {ButtonGroup, FormGroup} from "reactstrap";
+import {Typeahead} from "react-bootstrap-typeahead";
 
-const SorterFilter = () => (
-  <div className="sorter-select font-weight-bold px-3 py-2 bg-secondary rounded v-align-center my-3">
-    <span className="sort-icon bg-dark text-white rounded-circle d-inline-block text-center p-2 mr-2">
-      <FontAwesomeIcon icon={faSortAmountDown}/>
-    </span> Sort and filter <span className="float-right pt-1"><FontAwesomeIcon size="2x" icon={faAngleDown}/></span>
-  </div>
+const FilterMenu = ({open, close}) => (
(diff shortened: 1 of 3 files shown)


app/js/components/Buyer/SorterFilter.jsx
```diff
@@ -1,18 +1,91 @@
...
+          Card transfer
+        </CheckButton>
+      </ButtonGroup>
+
+      <h5 className="mt-4">Country</h5>
+      <FormGroup>
+        <Typeahead
+          options={['Canada', 'USA', 'France']}
+          placeholder={'Select'}
+        />
```
alaibe: I can see we will have multiple times this selector multiple times: create and list, maybe it is a component?

app/js/components/Buyer/SorterFilter.jsx
```diff
@@ -1,18 +1,91 @@
...
+      </FormGroup>
+
+      <h5 className="mt-4">Asset</h5>
+      <FormGroup>
+        <Typeahead
```
alaibe: I can see we will have multiple times this selector multiple times: create and list, maybe it is a component?

app/js/containers/HomeContainer.scss
```diff
@@ -8,6 +8,6 @@
     right: 0;
     bottom: 0;
     left: 0;
-    height: 7rem;
+    padding: 20px;
```
alaibe: I tend to prefer to use `rem` over `px` for sizing
What do you think?
Not valuable in this case as it is mobile only?

app/js/components/Buyer/SorterFilter.jsx
```diff
@@ -1,18 +1,91 @@
...
+import {Typeahead} from "react-bootstrap-typeahead";
 
-const SorterFilter = () => (
-  <div className="sorter-select font-weight-bold px-3 py-2 bg-secondary rounded v-align-center my-3">
-    <span className="sort-icon bg-dark text-white rounded-circle d-inline-block text-center p-2 mr-2">
-      <FontAwesomeIcon icon={faSortAmountDown}/>
-    </span> Sort and filter <span className="float-right pt-1"><FontAwesomeIcon size="2x" icon={faAngleDown}/></span>
-  </div>
+const FilterMenu = ({open, close}) => (
+  <Fragment>
```
alaibe: There is probably a good reason but why didn't you use modal from reactstrap?

jrainville: Really didn't think of using a modal to be honest. It never occurred to be that I could use them because of how they do not have the same shape, but I agree it could have worked, since it handles the backdrop for us. 

We will see if it bites us in the butt, but at least it gives us more control for now.

app/js/containers/HomeContainer.scss
jrainville: I used the values from Figma and they are only in `px`. I agree that `rem` are better though.

PR #59, opened 2019-02-09
feat: connect store and profile

alaibe:


app/js/features/metadata/reducer.js
```diff
@@ -0,0 +1,29 @@
...
+const DEFAULT_STATE = {
+  sellers: {},
+  offers: {}
+};
+
+function reducer(state = DEFAULT_STATE, action) {
+  switch (action.type) {
+    case LOAD_SELLER_SUCCEEDED:
+      return {
+        ...state, ...{
```
bgits: I don't think you need to spread the `sellers` object, it will eval to the same as 
```
{...state, sellers: { ...state.sellers, [action.address]: action.seller} }
```
but with an extra operation

alaibe: :+1: updated

app/js/features/metadata/saga.js
```diff
@@ -0,0 +1,53 @@
...
+import {
+  LOAD, LOAD_SELLER, LOAD_SELLER_FAILED, LOAD_SELLER_SUCCEEDED, LOAD_OFFERS_SUCCEEDED, LOAD_OFFERS_FAILED, LOAD_OFFERS
+} from './constants';
+
+export function *loadSeller({address}) {
+  try {
+    const size = yield MetadataStore.methods.sellersSize().call();
+    if (size === '0'){
+      return;
+    }
```
jrainville: Is the size check useful? Because knowing that the size is greater than 0 doesn't mean the address specified is a seller anyway.

alaibe: Interesting: the size check is annoying as if there is no seller, the following call fails
I could let it failed but it looked bad too so I made this extra call.

What do you think?

jrainville: Ah ok, thanks for clarifying. I did think it could be something like that. Maybe the array is not initiated if there is no seller. 
Anyway, it's no big deal, just wanted to know

PR #62, opened 2019-02-13
Feat/submit seller

alaibe:
* Submit Seller to Metadata Store
* Move License component
* Align MetadataStore naming and NewSeller naming
* Fix typo: Status Contract Code => Status Contact Code
* Update ProfileContainer Offers


app/js/components/ContactForm.jsx
```diff
@@ -25,13 +18,23 @@ class ContactForm extends Component {
...
           <FormGroup>
             <Label for="contactCode">Status contact code or Status ENS name</Label>
-            <Input type="text" name="contactCode" id="contactCode" value={contactCode}
-                   className="form-control" onChange={(e) => this.changeContactCode(e)}  validations={[required]}/>
+            <Input type="text"
+                   name="contactCode"
+                   id="contactCode"
+                   value={statusContactCode}
+                   className="form-control"
+                   onChange={(e) => this.props.changeStatusContactCode(e.target.value)}
```
jrainville: Don't you find it annoying to change the redux state for every character? That's why I used a local state and saved on page change. 
I know that it doesn't change much, but I'm not a fan of calling actions 20 times in a row.

app/js/components/Seller/MarginSelectorForm.test.jsx
```diff
@@ -5,7 +5,11 @@ import MarginSelectorForm from "./MarginSelectorForm";
 
 describe('MarginSelectorForm', () => {
   it('should render correctly', () => {
-    const component = shallow(<MarginSelectorForm margin={{}} fiat={{}} onMarginChange={() => undefined}/>);
+    const component = shallow(<MarginSelectorForm margin={0}
+                                                  margin={0}
```
jrainville: You put margin twice

app/js/components/Wizard.jsx
```diff
@@ -22,11 +22,15 @@ class Wizard extends Component {
...
+    return currentStep <= stepsLength - 1;
   };
 
+  lastStep() {
+    return this.state.currentStep === this.props.steps.length - 1;
+  }
+
   next = () => {
-    if (!this.canNext()) {
+    if (!this.canNext() || this.lastStep()) {
```
jrainville: Not sure why you added `lastStep`. If in `canNext` you just remove the `=` you added, you don't need `lastStep()`

app/js/containers/Seller/1_SellerLocationContainer.jsx
```diff
@@ -1,16 +1,17 @@
...
+
+class SellerLocationContainer extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      location: props.location
+      location: props.seller.location
     };
-    this.validate(props.location);
+    this.validate();
```
jrainville: Does this work? I know `location` is kind of global, but in the function, the arg will be undefined, no?

app/js/containers/Seller/3_SellerCurrencyContainer.jsx
```diff
@@ -17,53 +18,54 @@ const CURRENCY_DATA = [
...
     this.props.footer.onPageChange(() => {
-      props.setFiatCurrency(this.state.fiat);
+      props.setCurrency(this.state.currency);
     });
   }
 
-  validate(fiat) {
-    if (!fiat.id) {
+  validate(currency) {
+    if (!currency) {
```
jrainville: Is `currency` just a string now?

app/js/features/metadata/actions.js
```diff
@@ -1,3 +1,18 @@
-import { LOAD } from './constants';
+import { LOAD, ADD_SELLER } from './constants';
+import MetadataStore from 'Embark/contracts/MetadataStore';
 
 export const load = (address) => ({ type: LOAD, address });
+
+export const addSeller = (seller) => ({
```
jrainville: I would call this `addOffer` instead

app/js/components/ContactForm.jsx
alaibe: I don't call redux on each change, `this.props.changeStatusContactCode` is a function coming from the local state of the parent component

jrainville: Omg, sorry. I thought this was the container.

app/js/components/Wizard.jsx
alaibe: Maybe you have a more elegant way of doing that:

* canNext is used to know when we display or not the next button. For the final steps we want to be able to display the final button so canNext must be true on last step
* next is used to perform the increase but in case of last step there is no increase

app/js/containers/Seller/1_SellerLocationContainer.jsx
alaibe: This is a mistake but it seems to work...weird

app/js/containers/Seller/3_SellerCurrencyContainer.jsx
alaibe: yes

app/js/features/metadata/actions.js
alaibe: Agree

app/js/containers/Seller/3_SellerCurrencyContainer.jsx
jrainville: So we match the currency string with the string using the array above to get the symbol and other stuff I guess?

alaibe: Yes, the idea is that what is in local state will be in redux state and submitted later to the contract.

app/js/components/Wizard.jsx
jrainville: Aaah, ok, so `canNext` is used somewhere else to display a button, but we stop it here by checking if it's the last and if so, we don't actually go next. Do I understand correctly?

PR #68, opened 2019-02-15
Feat/misc

alaibe:


app/js/components/Profile/UpdateUser.jsx
```diff
@@ -0,0 +1,25 @@
+import React, {Component} from 'react';
+import {Row, Col, Button} from 'reactstrap';
+import PropTypes from 'prop-types';
+import {withNamespaces} from 'react-i18next';
+
+class UpdateUser extends Component {
```
jrainville: That's a weird component. I feel like it is never used alone, but might be wrong.
Also, it could be stateless

app/js/components/Seller/MarginSelectorForm.jsx
```diff
@@ -28,7 +28,7 @@ class MarginSelectorForm extends Component {
                      className="form-control prepend"
                      value={margin}
                      onChange={(e) => this.props.marginChange(e.target.value)}
-                     validations={[required, isNumber]} />
+                     validations={[required, isNumber, lowerThan.bind(null, 100)]} />
```
jrainville: It could be higher than 100, just that no one with any right mind would do it

app/js/containers/Seller/4_SellerMarginContainer.jsx
```diff
@@ -29,18 +29,19 @@ class SellerMarginContainer extends Component {
   }
 
   validate(margin) {
-    if (margin || margin === 0) {
+    if ((margin || margin === 0) && margin < 100) {
```
jrainville: Again, don't feel like it's a big issue

app/js/components/Seller/MarginSelectorForm.jsx
alaibe: Are you sure, that could avoid someone to make mistakes?

As you said, no one with any right mind would do it so why allow it?
Imagine I type 150 instead of 15

app/js/components/Profile/UpdateUser.jsx
alaibe: Yes Now that we have style for everything I will have a look to the components I can see this one being CenterButton or something similar, we have it in multiple place

jrainville: Yeah that makes more sense

app/js/components/Seller/MarginSelectorForm.jsx
jrainville: Yeah ok I'll let you have it. We probably want to be greater or equal than 0 too though.

alaibe: Yep, that makes sense, will add that

PR #71, opened 2019-02-18
feat: update balances when needed

alaibe:


app/js/components/YourSNTBalance.test.jsx
```diff
@@ -5,7 +5,7 @@ import YourSNTBalance from "./YourSNTBalance";
 
 describe('YourSNTBalance', () => {
   it('should render correctly', () => {
-    const component = shallow(<YourSNTBalance value={10}/>);
+    const component = shallow(<YourSNTBalance value={"10"}/>);
```
jrainville: Shouldn't the balance be a number?

alaibe: They are big number and as such represented as string

PR #72, opened 2019-02-18
feat: sell without license

alaibe:


app/js/layout/App.jsx
```diff
@@ -66,35 +68,41 @@ class App extends Component {
...
+              ]}/>
+            }
+
+            <Route path="/tmp/price" component={PriceContainer}/>
+            <Route path="/tmp/escrows" component={EscrowsContainer}/>
+            <Route path="/tmp/map" component={MapContainer}/>
+            <Route path="/tmp/signature" component={SignatureContainer}/>
+            <Route path="/tmp/arbitration" component={ArbitrationContainer}/>
+
+            <Redirect to="/"/>
```
jrainville: What does this do? Asking for my knowledge.

alaibe: This is a catch all in case the route is not known, redirect to home
We can also add a 404 page once we/if have it

jrainville: Ok, thanks for the answer. Indeed, a 404 page would be better. I'll add a Pivotal task

PR #73, opened 2019-02-18
A license cost is 10 SNT

alaibe:


app/js/features/license/saga.js
```diff
@@ -22,10 +22,6 @@ import {eventChannel} from "redux-saga";
 export function *doBuyLicense() {
   try {
     const price = yield call(License.methods.getPrice().call);
-    const balance = yield call(SNT.methods.balanceOf(web3.eth.defaultAccount).call);
-    if (balance < price) {
-      throw new Error(`Insufficient funds to buy a License. You need to have at least ${price} SNT, but you have ${balance}`);
-    }
```
alaibe: The UI already prevent that

app/js/features/license/saga.js
```diff
@@ -22,10 +22,6 @@ import {eventChannel} from "redux-saga";
 export function *doBuyLicense() {
```
alaibe: The next stage would be to pass the price so we could use the normal doTransaction

app/js/features/license/saga.js
jrainville: The UI can bug or be bypassed by someone trying to "cheat".
More validation is never bad IMO. Unless you think it slows down the process too much

alaibe: The contract will always have your back, and that will result in an error too no?

alaibe: > The UI can bug

Never seen one of those!!

jrainville: It does, but the response from the contract is sometimes very cryptic, especially when using `estimateGas` (in that case the error in only "always failing transaction")

alaibe: If we really want to go down that route it means we triple all validation:
1) Real validation are in solidity
2) The UI offers validation by not allowing submit
3) The saga offers validation

IMO a combination of 1 and 2 are enough, as the only way to go through 3) is as you said: "cheating" by manipulating the code or a bug:
*Cheating: We cannot prevent that as you can also remove the validation in saga too, I wouldn't try to fix that case as it is impossible
*Bug: This is of course unplanned and you are dealing only with 1 case here, what about all others:
```
require(licenseDetails[_owner].creationTime == 0, LICENSE_ALREADY_BOUGHT);
require(token.allowance(_owner, address(this)) >= price, "Allowance not set for this contract to expected price");
require(token.transferFrom(_owner, address(this), price), UNSUCCESSFUL_TOKEN_TRANSFER);
```
At the end, there is little chance your bug is where you think it was in your guard

Also, there is some chances that more will be added to solidity but client will be untouched, that is why we should deal with that in a generic manner for the end user (transaction error page, with retry?)
And for us, find a way to report the error (geth don't help)

> More validation is never bad IMO

Probably another subject but I disagree (I'm french :joy: )
If you have 2 validations about the same thing in 2 different place, there is a lot of chance the next developer changing validation will blame you => The second one is dead code and can never be reached

PR #74, opened 2019-02-19
feat: allow removal of offers

richard-ramos:


test/metadata_store_spec.js
```diff
@@ -99,4 +99,19 @@ contract("MetadataStore", function () {
...
+    const offerId = offerAdded.returnValues.offerId;
+
+    const receipt2 = await MetadataStore.methods.removeOffer(offerId).send();
+    const offerRemoved = receipt2.events.OfferRemoved;
+    assert(!!offerRemoved, "OfferRemoved() not triggered");
+    assert.equal(offerRemoved.returnValues.owner, accounts[0], "Invalid seller");
+    assert.equal(offerRemoved.returnValues.offerId, offerId, "Invalid offer");
+
+    const offer = await MetadataStore.methods.offers(offerId).call();
+console.log(offer);
```
alaibe: Probably a left over?

contracts/teller-network/MetadataStore.sol
```diff
@@ -187,6 +192,22 @@ contract MetadataStore is Ownable {
...
+    /**
+    * @notice Remove user offer
+    * @param _offerId Id of the offer to remove
+    */
+    function removeOffer(
+        uint256 _offerId
+    ) public {
+        require(userWhitelist[msg.sender], "User does not exist");
+        require(offerWhitelist[msg.sender][_offerId], "Offer does not exist");
+        require(offers[_offerId].status == OfferStatus.Open, "Only open offers can be removed");
```
alaibe: Is the state machine really usefull? if not then we can remove it

test/metadata_store_spec.js
richard-ramos: oops!

contracts/teller-network/MetadataStore.sol
richard-ramos: I'll remove it. We could always introduce it again in the future if necessary.

PR #75, opened 2019-02-19
Feat/offers list

alaibe:


stories/offers.stories.jsx
```diff
@@ -14,9 +14,8 @@ storiesOf('Offers', module)
           {
             asset: '0x0',
             currency: 'EUR',
-            paymentMethods: ['Credit Card'],
-            margin: 1,
-            marketType: 1
+            paymentMethodsForHuman: ['Credit Card'],
```
richard-ramos: Is Credit Card a valid payment type?

alaibe: Updated, it is not

PR #77, opened 2019-02-19
feat: load price for currency and apply margin calculation in MarginSelectorForm

richard-ramos:


app/js/containers/Seller/4_SellerMarginContainer.jsx
```diff
@@ -5,6 +5,10 @@ import {connect} from 'react-redux';
 import MarginSelectorForm from '../../components/Seller/MarginSelectorForm';
 import Loading from '../../components/ui/Loading';
 import newSeller from "../../features/newSeller";
+import network from '../../features/network';
+import prices from '../../features/prices';
+import {getPrices} from '../../features/prices/reducer';
```
alaibe: You should not be able to access internal element of each feature, that should exposed through the index.
You can see example for others

app/js/containers/Seller/4_SellerMarginContainer.jsx
```diff
@@ -26,6 +30,11 @@ class SellerMarginContainer extends Component {
     } else {
       this.setState({ready: true});
     }
+
+    this.props.fetchPrices({
+      from: [this._getSymbol()],
+      to: [this.props.seller.currency]
```
jrainville: Why are those arrays?

app/js/containers/Seller/4_SellerMarginContainer.jsx
```diff
@@ -26,6 +30,11 @@ class SellerMarginContainer extends Component {
     } else {
       this.setState({ready: true});
     }
+
+    this.props.fetchPrices({
+      from: [this._getSymbol()],
```
alaibe: The problem with this solution is that the price changes all the time and if you left the page open for 1 hours the data displayed might not be accurate.

What about when we do the initialisation we also start a scheduler who is going to fetch prices for all token/currency periodically => e.g every 1 or n minutes.

We can then attach the prices to the different tokens we have

What do you think?

app/js/containers/Seller/4_SellerMarginContainer.jsx
richard-ramos: fetchPrice allows asking for the price for more than a single token symbol.  In this case _getSymbol() returns the seller selected asset, and currency, the fiat currency symbol (EUR, USD...)

app/js/containers/Seller/4_SellerMarginContainer.jsx
```diff
@@ -48,13 +57,20 @@ class SellerMarginContainer extends Component {
     this.setState({marketType});
   };
 
+  _getSymbol(){
+    return Object.keys(this.props.tokens)
+                 .find(token => this.props.tokens[token].address === this.props.seller.asset);
```
alaibe: you could do a selector for that I think

app/js/containers/Seller/4_SellerMarginContainer.jsx
jrainville: If we do that, don't forget to kill the interval on component unmount

alaibe: I would not even do that :)
I would imagine the token always have a set of price attach to them that is up to date

So not related to the specific component but to the App

jrainville: Ah ok, yeah, if the fetch is in the App, it's ok then

app/js/features/prices/saga.js
```diff
@@ -16,4 +18,23 @@ export function *onFetchPrices() {
   yield takeEvery(FETCH_PRICES, doFetchPrices);
 }
 
-export default [fork(onFetchPrices)];
+
+function *fetchPricesPeriodically() {
```
jrainville: No need to put "periodically" in the function name, since the function itself doesn't do an interval

PR #78, opened 2019-02-19
Avoid to re-render component

alaibe:


app/js/layout/App.jsx
```diff
@@ -59,6 +60,11 @@ class App extends Component {
     }
   }
 
+  shouldComponentUpdate(prevProps) {
+    return prevProps.isReady !== this.props.isReady ||
+      !_.isEqual(prevProps.profile, this.props.profile);
```
jrainville: Shouldn't you just have to move this to the `componentDidUpdate` `if`?

alaibe: Not sure to see what you mean

PR #79, opened 2019-02-19
Enable specifying ENS names as contact codes

jrainville:
I had to add redux-thunk, because getting the ENS address in the action is way easier. Let me know if I shouldn't or there is a better way, but I couldn't find a good way to to it in the saga (especially since we use `doTransaction`).

### DIFF
--- a/app/js/containers/EditProfileContainer.jsx
+++ b/app/js/containers/EditProfileContainer.jsx
@@ -1,4 +1,4 @@
-import React, {Component} from 'react';
+import React, {Component, Fragment} from 'react';
 import {withRouter} from "react-router-dom";
 import PropTypes from 'prop-types';
 import {connect} from "react-redux";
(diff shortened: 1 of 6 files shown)


jrainville:
@alaibe I don't mind using saga, but I really don't know an easy way to do that. Like I said in the description, since we sue `doTransaction`, it's really hard to actually modify the payload before sending.

Maybe we can pair to find a way. Otherwise, I feel like the `thunk` solution really is not that bad.

app/js/utils/saga.js
```diff
@@ -44,3 +45,22 @@ export function *doTransaction(preSuccess, success, failed, {value = 0, toSend})
     yield put({type: failed, error: error.message});
   }
 }
+
+export function getEnsAddress(name) {
```
alaibe: there is an embark js service, that would probably be a better location

alaibe: We probably need to update the contract to allow non address?

alaibe:
@jrainville see my latest commit on this PR and tell me if it works.
I have nothing against  `thunk`  and I am fine moving to it if you feel it is simpler. 
Having the 2 together would be like having `underscore` and `lodash` => bigger build...

jrainville:
You are the best @alaibe, your solution is very good!

https://i.redd.it/fskhv6es09v01.png

jrainville:
Updated without Thunk!

app/js/utils/saga.js
jrainville: Moved. Also, update the contract to be a string instead of an address?

PR #80, opened 2019-02-19
fix(profile): compact address and show it hover

jrainville:
image


app/js/components/SellerInformation.jsx
```diff
@@ -1,25 +1,44 @@
...
+    return (<Row className="border rounded py-4 m-0 text-center shadow-sm">
+      <Col xs="12">
+        <Blockies seed={address} className="rounded-circle border" scale={8}/>
+      </Col>
+      <Col xs="12">
+        <h4 className="font-weight-bold">{name}</h4>
+      </Col>
+      <Col xs="12">
+        <p className="text-muted" onMouseOver={this.mouseOverAddress} onMouseOut={this.mouseOutAddress}>
+          {this.state.addressHovered ? address : compactAddress(address)}
```
alaibe: There is an Address component do you think it is possible to use it?

jrainville: Oh right. I'll add the state to that component for the hover

jrainville: Done

PR #88, opened 2019-02-20
fix: initial price load

richard-ramos:


app/js/features/network/saga.js
```diff
@@ -14,6 +15,7 @@ export function *doInit() {
     yield call(onReady);
     const networkId = yield call(web3.eth.net.getId);
     yield put({type: INIT_SUCCEEDED, networkId});
+    yield put({type: PRICE_INTERVAL});
```
jrainville: I'd rename the constant to `FETCH_PRICES` because it looks like the constant contains the amount of time in the interval

richard-ramos: is `FETCH_EXCHANGE_RATE` ok?

jrainville: Yeah

PR #89, opened 2019-02-20
feat: show access button only on status

richard-ramos:


app/js/components/ContactForm.jsx
```diff
@@ -35,7 +36,7 @@ class ContactForm extends Component {
                    className="form-control"
                    onChange={(e) => this.props.changeStatusContactCode(e.target.value)}
                    validations={[required]}/>
-            <Button className="input-icon p-0" color="link">Give access</Button>
+            {web3.currentProvider.status && <Button className="input-icon p-0" color="link">Give access</Button>}
```
alaibe: I'd avoid to use global especially in presentation components.
I think the provider and isStatus should be in global state?

What do you think @jrainville @richard-ramos

jrainville: Ah yeah, you surely are gonna have problems in the tests and storybook. 
Also, I do agree that the presentation component shouldn't have to do that

richard-ramos: I guess we'll have to do that for web3.eth.defaultAccount as well since it's used in EscrowList and ArbitrationList?

alaibe: default account is already there, you can see here:
https://github.com/status-im/status-teller-network/blob/master/app/js/features/network/reducer.js#L11

You also have information about network id and name

richard-ramos: nice. I'll add the provider there too!. Thx

PR #93, opened 2019-02-20
feat: add error page and no provider message

jrainville:
image


app/js/features/network/selectors.js
```diff
@@ -1,4 +1,14 @@
 export const isReady = state => state.network.ready;
+export const error = state => state.network.error;
+export const errorTip = state => {
```
alaibe: Too late because of merged, but you bind component and logic too much here probably
I would have named that: `noProviderError: boolean`.

Also because that way i18n won't work well :)

app/js/features/network/reducer.js
```diff
@@ -36,6 +33,13 @@ function reducer(state = DEFAULT_STATE, action) {
         tokens
       };
     }
+    case INIT_FAILED: {
+      return {
+        ...state,
+        ready: false,
+        error: action.error
```
alaibe: error needs to be reset when init success

PR #100, opened 2019-02-25
fix: skip ens lookup if contact code is specified

richard-ramos:


app/js/features/metadata/saga.js
```diff
@@ -86,7 +87,10 @@ export function *onAddOffer() {
 }
 
 export function *updateUser({user}) {
-  user.statusContactCode = yield getEnsAddress(user.statusContactCode);
+  if (!contactCodeRegExp.test(user.statusContactCode)) {
```
jrainville: This condition should be moved to here: https://github.com/status-im/status-teller-network/blob/master/app/js/services/embarkjs.js#L19

I had already added a check for addresses, but forgot that contact codes are not addresses.

PR #106, opened 2019-02-25
feat: add trade to contract

alaibe:


richard-ramos:
@alaibe , 
See  https://github.com/status-im/status-teller-network/compare/feat/trade-escrow?expand=1 
We might be able to avoid storing the trades as a separate entity from the escrows.  What do you think?

PR #107, opened 2019-02-26
feat: buyers and sellers can initiate escrow

richard-ramos:


contracts/teller-network/Escrow.sol
```diff
@@ -25,22 +28,28 @@ contract Escrow is Pausable, MessageSigned {
...
-        bool released;
-        bool canceled;
-        bool paid;
         uint rating;
+        uint256 tradeAmount;
+        TradeType tradeType;
+        EscrowStatus status;
     }
 
+    enum TradeType {FIAT, CRYPTO}
```
alaibe: Currently there is only 1 trade type right?

contracts/teller-network/Escrow.sol
```diff
@@ -61,43 +70,113 @@ contract Escrow is Pausable, MessageSigned {
...
+     * @param _seller Seller address
+     * @param _tradeAmount Amount buyer is willing to trade
+     * @param _tradeType Indicates if the amount is in crypto or fiat
      * @param _token Token address. Must be 0 for ETH
+     * @dev Requires contract to be unpaused.
+     *         The seller needs to be licensed.
+     */
+    function create(address payable _buyer, address payable _seller, address _token, uint _tradeAmount, uint8 _tradeType) public whenNotPaused {
+        require(msg.sender == _buyer || msg.sender == _seller, "Must participate in the trade");
+        require(license.isLicenseOwner(_seller), "Must be a valid seller to create escrow transactions");
```
alaibe: Should we also verify that the offer exists?

contracts/teller-network/Escrow.sol
```diff
@@ -61,43 +70,113 @@ contract Escrow is Pausable, MessageSigned {
...
      * @dev Requires contract to be unpaused.
      *         The seller needs to be licensed.
      *         The expiration time must be at least 10min in the future
      *         For eth transfer, _amount must be equals to msg.value, for token transfer, requires an allowance and transfer valid for _amount
      */
-    function create(address payable _buyer, uint _amount, address _token, uint _expirationTime) public payable whenNotPaused {
+    function fund(uint _escrowId, uint _tokenAmount, uint _expirationTime) public payable whenNotPaused {
+        require(_escrowId < transactions.length, INVALID_ESCROW_ID);
         require(_expirationTime > (block.timestamp + 600), "Expiration time must be at least 10min in the future");
         require(license.isLicenseOwner(msg.sender), "Must be a valid seller to create escrow transactions");
```
alaibe: You don't verify that escrow seller and msg sender match?

contracts/teller-network/Escrow.sol
```diff
@@ -61,43 +70,113 @@ contract Escrow is Pausable, MessageSigned {
...
+     * @param _token Token address. Must be 0 for ETH
+     * @param _tokenAmount How much ether/tokens will be put in escrow
+     * @param _expirationTime Unix timestamp before the transaction is considered expired
+     * @dev Requires contract to be unpaused.
+     *         The seller needs to be licensed.
+     *         The expiration time must be at least 10min in the future
+     *         For eth transfer, _amount must be equals to msg.value, for token transfer, requires an allowance and transfer valid for _amount
+     */
+    function create_and_fund(address payable _buyer, address _token, uint _tokenAmount, uint _expirationTime, uint _tradeAmount, uint8 _tradeType) public payable whenNotPaused {
+        require(license.isLicenseOwner(msg.sender), "Must be a valid seller to create and fund escrow transactions");
```
alaibe: Should this function just called the 2 underlying function?
That will ensure validation are done once and correctly no?

contracts/teller-network/Escrow.sol
```diff
@@ -223,15 +301,17 @@ contract Escrow is Pausable, MessageSigned {
...
      */
     function _cancel(uint _escrowId, EscrowTransaction storage trx) private {
-        trx.canceled = true;
-
-        if(trx.token == address(0)){
-            trx.seller.transfer(trx.amount);
-        } else {
-            ERC20Token token = ERC20Token(trx.token);
-            require(token.transfer(trx.seller, trx.amount));
+        if(trx.status == EscrowStatus.FUNDED){
```
alaibe: There is probably some state where it is not possible to cancel?

contracts/teller-network/Escrow.sol
```diff
@@ -244,16 +324,33 @@ contract Escrow is Pausable, MessageSigned {
...
-
-        require(!trx.released, TRANSACTION_ALREADY_RELEASED);
-        require(!trx.canceled, TRANSACTION_ALREADY_CANCELED);
-        require(!trx.paid, "Cannot withdraw an already paid transaction. Open a case");
-
+        require(trx.status == EscrowStatus.FUNDED, "Cannot withdraw from escrow in a stage different from FUNDED. Open a case");
         _cancel(_escrowId, trx);
     }
 
+
```
alaibe: seems like a lot of spaces :joy:

contracts/teller-network/Escrow.sol
jrainville: We could, but it's an additional layer of complexity. Also, this way let's the seller negotiate and do whatever escrow he wants. 
In the bad case where the buyer "cheats" and puts a lower price, the seller can just not fund the escrow.

alaibe: fair enough

contracts/teller-network/Escrow.sol
richard-ramos: you can never have too much space! :)

contracts/teller-network/Escrow.sol
richard-ramos: I was thinking of using the tradeType and tradeAmount to store the information from this screen. Maybe we can use a string data type instead of those two fields and store the info we need directly on that string?  (to give an example, If i want to open a trade for 500USD, the field would have a string containing `FIAT:500:USD`, or if the trade is for 12500SNT,  `CRYPTO:12500:SNT`. 

screenshot

contracts/teller-network/Escrow.sol
richard-ramos: I imagine a cancel could happen only for a `CREATED` or `FUNDED` state, with `FUNDED` being the only state where there are eth/tokens stored in the contract.

contracts/teller-network/Escrow.sol
richard-ramos: Now we do in line 121

contracts/teller-network/Escrow.sol
richard-ramos: An additional change that came from this observation is that a seller can now call `fund` any number of times to keep adding crypto to the tokens while the escrow status is CREATED or FUNDED.

Do you think the seller should be able to keep funding the escrow in any status different from CANCELED or RELEASED?

PR #108, opened 2019-02-27
feat: create escrow page

alaibe:


app/js/pages/Escrow/index.scss
```diff
@@ -0,0 +1,7 @@
+@import "../../../css//variable-overrides.scss";
+
+.escrow {
+  .escrow--split {
+    background-color: $secondary;
```
jrainville: There is `.bg-secondary` that you can use

app/js/pages/Profile/index.jsx
```diff
@@ -29,7 +31,16 @@ class Profile extends Component {
         <h3 className="mt-3">{profile.location}</h3>
         <Map coords={{latitude: 45.492611, longitude: -73.617959}} markerOnly={true}/>
         <p className="text-muted mt-2">Saalestraße 39A, 12055 Berlin</p>
-        <Offers offers={profile.offers} onClick={this.offerClick}/>
+        <Row>
+          <Col xs="12" className="mt-2">
```
jrainville: IMO, doing a row just for one col doesn't do much, it's the same as a normal div at this point. Am I missing something?

app/js/ui/RoundedIcon/index.scss
```diff
@@ -0,0 +1,19 @@
+@import "../../../css/variable-overrides.scss";
+
+.rounded-icon {
+  padding: 10px;
+}
+
+.rounded-icon__grey {
+  background-color: #939BA1;
```
jrainville: If the color is not in the variables, you can add it. We will probably reuse it.

app/js/pages/Profile/index.jsx
alaibe: Yes, Row are removing the padding from the container, so the background is full width

jrainville: Ah ok, understood

app/js/pages/Escrow/index.scss
alaibe: updated

app/js/ui/RoundedIcon/index.scss
alaibe: updated, it was but I missed it!!

PR #112, opened 2019-03-01
Add user when creating escrow

alaibe:


contracts/teller-network/Escrow.sol
```diff
@@ -146,13 +157,26 @@ contract Escrow is Pausable, MessageSigned {
...
+    function create_and_fund(
+        address payable _buyer,
+        address _token,
+        uint _tokenAmount,
+        uint _expirationTime,
+        uint _tradeAmount,
+        uint8 _tradeType,
+        bytes memory _statusContactCode,
+        string memory _location,
+        string memory _username
```
jrainville: Buyer information won't be available here, no? Because this function is only called by the seller. Unless the seller gets the info for the buyer? Hmm, could make sense actually since they probably complete it together, but it's still a bit useless, since the username and contact code is only useful for distant comms.

What do you think?

alaibe: According to what we have now, this is the buyer creating the escrow and the seller funding it.

If a seller need to create an escrow then he will need information of the seller but it is not in our flow right now or I miss it.

I think we should do first minimum in order to get in work according to the flow we have and making sure we don't have problem with that already.
One problem I can see is that someone could change someone else username/location, which is a bit worrying me at the stage

jrainville: Well, my comment was specifically for `create_and_fund`, because that really is for the seller to call. 
And your point about changing another user really enforces my view that the seller shouldn't set the username and location of the buyer.

So, for `create`, the one that buyer calls, it's fine to set its infos, but `create_and_fund` should be only about the escrow

alaibe: I almost removed `create_and_fund` as it is not part of our use-cases
What do you think?

jrainville: Well, it's useful for when the buyer doesn't have ETH and can't create, so goes to a local seller and the seller creates the escrow for them.

I know we don't have a design for that yet, but I feel it can be really useful

PR #116, opened 2019-03-04
Add location filter

jrainville:
The way the filter works is, you enter a location, which can be an address, a city a postal code. Then, the system gets the coordinates of that location. We then calculate the distance between the location the user entered and the location of the sellers

We probably will want a check on the cache to not get the users/offers mutliple times in the future, because with hundreds of users, we will make 100s of requests for coordinates.


app/js/utils/location.js
```diff
@@ -0,0 +1,21 @@
+import {API_KEY} from "../features/metadata/constants";
```
alaibe: That could be a service:
googleMap.js

I feel the API_KEY could be there too

app/js/features/metadata/saga.js
```diff
@@ -16,6 +20,8 @@ export function *loadUser({address}) {
     }
     const id = yield MetadataStore.methods.addressToUser(address).call();
     const user = yield MetadataStore.methods.users(id).call();
+
+    yield put({type: LOAD_USER_LOCATION, user, address});
```
alaibe: 💯

app/js/features/metadata/saga.js
```diff
@@ -27,6 +33,20 @@ export function *onLoadUser() {
   yield takeEvery(LOAD_USER, loadUser);
 }
 
+export function *loadLocation({user, address}) {
+  try {
+    const location = yield getLocation(user.location);
```
alaibe: `const coords`, so you can use a shortcut on line 39

app/js/utils/location.js
```diff
@@ -0,0 +1,21 @@
...
+
+const geocoder = require('google-geocoder');
+
+export function getLocation(place) {
+  return new Promise((resolve, reject) => {
+    const geo = geocoder({
+      key: API_KEY
+    });
```
alaibe: Not sure if really valuable but that doesn't need to be instantiate all the time and that could be out of the function?

app/js/pages/OffersList/index.jsx
```diff
@@ -50,6 +55,34 @@ class OffersList extends Component {
...
+      this.setState({
+        calculatingLocation: false,
+        error: e.message,
+        location
+      });
+    });
+  };
+
+  calculateDistance = (userCoords) => {
+    return  Math.sqrt(Math.pow(userCoords.lat - this.state.locationCoords.lat, 2) + Math.pow(userCoords.lng - this.state.locationCoords.lng, 2));
```
alaibe: 👍 for the formula
👎 for the extra space between return and Math

Also should it be in utils?

PR #118, opened 2019-03-05
Remove seller and token address from escrows and add offer Id

richard-ramos:


contracts/teller-network/MetadataStore.sol
```diff
@@ -235,6 +237,7 @@ contract MetadataStore is Ownable {
         );
     }
 
+
```
jrainville:
```suggestion

```

PR #122, opened 2019-03-06
Real rating

jrainville:
- Fix/move getting the rating to the escrows saga
- Add escrows and ratings in the data (afterDeploy)
 - !!! I need help here. I had to comment out the `payFee` function in the contract. That's the thing that's blocking me putting it as a real PR
- Add the rating sort

image


app/js/features/escrow/saga.js
```diff
@@ -120,7 +121,55 @@ export function *onLoadEscrows() {
   yield takeEvery(LOAD_ESCROWS, doLoadEscrows);
 }
 
+
+export function *checkUserRating({address}) {
+  try {
+    address = address || web3.eth.defaultAccount;
+    const state = yield select();
+    const offers = Object.values(state.metadata.offers).filter(offer => offer.owner.toLowerCase() === address.toLowerCase());
```
alaibe: all address on the store should be lowerCase.
we do this pre-processing when we store them in the reducer

app/js/features/escrow/saga.js
```diff
@@ -120,7 +121,55 @@ export function *onLoadEscrows() {
...
+    if (!offers.length) {
+      return yield put({type: USER_RATING_SUCCEEDED, userRating: -1, address});
```
alaibe: in case of no offer should it be
```
downCount: 0, upCount: 0, voteCount: 0
```

app/js/features/metadata/reducer.js
```diff
@@ -85,6 +86,16 @@ function reducer(state = DEFAULT_STATE, action) {
         ...state, offers: newOffers
       };
     }
+    case USER_RATING_SUCCEEDED:
```
alaibe: feels weird to reply to an event from escrow in this specific reducer,
I had to do the same with escrow and offer
=> I leave the reducer as they were but move the mix in the reducer.

I feels this is because we have the wrong structure. I am not sure what it should be to avoid those mixed.
Any idea?

app/js/pages/OffersList/index.jsx
```diff
@@ -104,23 +98,25 @@ class OffersList extends Component {
...
       ));
       return grouped;
     }, {});
 
+    // Sort
+    let sortFunction;
+    switch (this.state.sortType) {
+      case 0: sortFunction = sortByRating; break;
+      case 1: sortFunction = sortByDate; break;
+      default: sortFunction = sortByRating;
```
alaibe: you only need 1 case and 1 default?

app/js/utils/sorters.js
```diff
@@ -0,0 +1,20 @@
+export function weightedScore(user) {
+  const upvotes = user.upCount + 4;
+  const downvotes = user.downCount + 4;
+  return upvotes / (upvotes + downvotes);
```
alaibe: out of curiosity, why the `+ 4`

config/data.js
```diff
@@ -1,6 +1,4 @@
-const LICENSE_PRICE = "10000000000000000000"; // 10 * Math.pow(10, 18)
```
alaibe: :100:

config/data.js
```diff
@@ -47,22 +45,49 @@ module.exports = async (deps) => {
     const currencies = ['USD', 'EUR'];
     const marketTypes = [0, 1];
 
-    await Promise.all(addresses.slice(1, 5).map(async (address) => {
+    const offerReceipts = await Promise.all(addresses.slice(1, 5).map(async (address) => {
       const addOffer = deps.contracts.MetadataStore.methods.addOffer(
-        tokens[Math.floor(Math.random()*tokens.length)],
+        tokens[1],
+        // TODO un hardcode token and check below
+        // tokens[Math.floor(Math.random() * tokens.length)],
```
alaibe: what is wrong with:
```
tokens[Math.floor(Math.random() * tokens.length)],
```
?

config/data.js
```diff
@@ -47,22 +45,49 @@ module.exports = async (deps) => {
...
     }));
 
+    console.log('Creating escrows and rating them...');
+    const expirationTime = parseInt((new Date()).getTime() / 1000, 10) + 10000;
+    const FIAT = 0;
+    const val = 1000;
+
+    const buyerAddress = addresses[1];
+    await Promise.all(addresses.slice(2, 5).map(async (creatorAddress, idx) => {
+      const ethOfferId = offerReceipts[idx - 1 + 2].events.OfferAdded.returnValues.offerId;
```
alaibe: To know offer id you just have to use the offerReceipts length -1, that would give you all offer ids

app/js/features/escrow/saga.js
jrainville: Oops, my bad, that's the remnant of an old rating calculation I had

app/js/features/metadata/reducer.js
jrainville: Not sure. We either have escrow function calls in the metadata saga or have escrow constants in the metadata reducer. There is no good way around it I think. 
I'd leave it like that for now as we don,t have a solution and I don't want to move stuff around just for the sake of it.

app/js/utils/sorters.js
jrainville: My way of implementing a Bayesian weighted average (https://en.wikipedia.org/wiki/Bayesian_average)

Basically, it's the predicted amount of likes and dislikes, so it balances the fact that a user with 1 like and 0 dislikes shouldn't be higher than someone with 10 likes and 1 dislike.

Tell me if it's not clear and also if maybe I did it wrong. My Bayes classes were a long time ago

config/data.js
jrainville: Nothing wrong, it's just that creating a releasing escrows with tokens is way harder because we have to approve the SNT allowance. So I wanted to keep it simple.

app/js/features/metadata/reducer.js
alaibe: Sure I was just asking to know if you have a better idea but you are like me :joy:

app/js/utils/sorters.js
alaibe: Humm it is not really the formula, also I understand what you are doing.
To make a mapping: C == 4, m == 1 and sum == upCount or downCount == n
Correct?

jrainville: Yeah something like that. 
I more specifically followed this: http://julesjacobs.github.io/2015/08/17/bayesian-scoring-of-ratings.html

PR #131, opened 2019-03-18
update to embark beta.1

iurimatias:


michaelsbradleyjr:
The project already has a `yarn.lock`, it shouldn't also have a `package-lock.json`, which is introduced by this PR.

emizzle:
> The project already has a `yarn.lock`, I don't believe it should also have a `package-lock.json`, which is introduced by this PR.

great catch, +1

PR #141, opened 2019-03-25
add hardSet state reconciler

bgits:


bgits:
related to: https://www.pivotaltracker.com/n/projects/2231548/stories/164274994

PR #156, opened 2019-04-02
 feat: display SNT fee when creating an offer

richard-ramos:


src/js/wizards/Sell/5_Margin/components/MarginSelectorForm.jsx
```diff
@@ -54,10 +55,12 @@ class MarginSelectorForm extends Component {
         </div>
         <small>{t('marginSelectorForm.priceOrigin')}</small>
 
-        <h3>{t('marginSelectorForm.ourFee')}</h3>
-        <div className="border rounded p-3">
-          TODO SNT
-        </div>
+        {fee !== '0' && <Fragment>
```
jrainville: From the container, it looks like while we are fetching the fee, it would be undefined while not resolved yet. 
Maybe here we should check that the fee is not undefined too?

PR #158, opened 2019-04-03
feat: set tmp logo

richard-ramos:


src/js/layout/Header/index.scss
```diff
@@ -5,4 +5,10 @@ header .navbar {
     font-size: 17px;
     font-weight: bold;
   }
+
+  img {
```
jrainville: You can use attributes on the image for that:
```
<img src={logo} alt="Logo" className="mr-1" width="32" height="32" />
```

PR #159, opened 2019-04-03
feat: confirm dialog

richard-ramos:


src/css/bootstrap-overrides.scss
```diff
@@ -1,6 +1,11 @@
 @import "variable-overrides";
 @import "../../node_modules/bootstrap/scss/bootstrap";
 
+#root {
+  width: 100%;
+  position: absolute;
```
jrainville: Why? Wouldn't relative work if you just need it for other absolutes?

src/js/components/ConfirmDialog/index.jsx
```diff
@@ -0,0 +1,42 @@
+import React, {Component} from 'react';
+import {Modal, ModalHeader, ModalBody, ModalFooter, Button} from 'reactstrap';
+import PropTypes from 'prop-types';
+
+class ConfirmDialog extends Component {
+
+  onContinue = () => {
+    this.props.onConfirm();
+    this.props.onCancel();
```
jrainville: Shouldn't this be only confirm?

richard-ramos: I used the onCancel() there to close the dialog right after calling the onConfirm.

jrainville: Ah ok. You shouldn't do that there, as the dialog itself doesn't know what `onConfirm` and `onCancel` do. 

Example, I could create a component that uses the dialog and on confirm, send a TX, but on cancel, switch page. Now that would be confusing because it would do both.

What you should do is juste handle the close in the other component (you can in that other component do both confirm and cancel if you like because it's the component itself that handles it and it knows that cancel only closes)

src/css/bootstrap-overrides.scss
richard-ramos: ah! I did that because the page changed width when the dialog was shown!
look: 

screencapture-localhost-3000-2019-04-04-08_51_05

jrainville: Oh wow. Ok then, if absolute fixes it then it's fine. (maybe make sure that the Sorter menu still works)

PR #171, opened 2019-04-19
fix: header styles and width

richard-ramos:
Adds a border-bottom and expands the width of the clickable area of the logo

### DIFF
--- a/src/css/bootstrap-overrides.scss
+++ b/src/css/bootstrap-overrides.scss
@@ -10,7 +10,6 @@
 header {
   padding-left: 15px;
   padding-right: 15px;
-  border-bottom: 1px solid #EEF2F5;
   margin-bottom: 20px;
   height: 65px;
   width: 100%;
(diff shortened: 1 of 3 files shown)


src/css/bootstrap-overrides.scss
```diff
@@ -7,6 +7,27 @@
   position: absolute;
 }
 
+header {
+  padding-left: 15px;
+  padding-right: 15px;
+  border-bottom: 1px solid #EEF2F5;
```
jrainville: Instead of adding the border here, you can use the bootstrap class `border-bottom`. It will add the border with the same style as everywhere so you don't have to get the color yourself. (https://getbootstrap.com/docs/4.0/utilities/borders/#border)

src/css/bootstrap-overrides.scss
```diff
@@ -7,6 +7,27 @@
...
+}
+
+.body-content {
+  padding-right: 15px;
+  padding-left: 15px;
+}
+
+.container {
+  width: 100%;
+  padding: 0;
```
jrainville: If you want, there is `.p-0` that odes that from bootstrap.

PR #172, opened 2019-04-19
fix: adds fiat value to offers, truncated to a max of 2 decimal places

richard-ramos:


src/js/components/Offer/index.scss
```diff
@@ -0,0 +1,4 @@
+.price {
+  color: #000;
+  font-weight: 400;
```
jrainville: There is `.font-weight-normal` for that.
For the color, I'm surpisd that there is none for `black`. What the color that you are overwriting here? (color: https://getbootstrap.com/docs/4.0/utilities/colors/)

src/js/components/Offer/index.jsx
```diff
@@ -5,8 +5,18 @@ import {Link} from "react-router-dom";
...
+
+function truncateTwo(num) {   
+  var re = new RegExp('^-?\\d+(?:\.\\d{0,' + (2 || -1) + '})?');
+  num = Number(num.toString().match(re)[0]);
+  return Number(Math.round(num + "e+2")  + "e-2");
+}
+
+const Offer = ({offer, withDetail, prices}) => {
+  const price = prices[offer.token.symbol][offer.currency];
+  const calcPrice = offer.marketType === "0" ? price * ((100 + (parseFloat(offer.margin))) / 100) : price * parseFloat(offer.margin) / 100;
```
jrainville: Lik I said in the other, might be better off as a utils

src/js/components/Offer/index.scss
richard-ramos: The color for links

src/js/components/Offer/index.jsx
richard-ramos: I'll extract this as utils in the other PR since it's more recent!

PR #173, opened 2019-04-19
fix: show fiat value in seller's profile offers when accessed by buyers

richard-ramos:


src/js/pages/Profile/components/Offer.jsx
```diff
@@ -1,23 +1,27 @@
...
+import {truncateTwo} from '../../../utils/numbers';
 import {TokenImages} from '../../../utils/images';
 
-const Offer = ({offer, onClick}) => (
-  <Row className="border py-2 mx-0 my-2 rounded">
-    <Col xs="6" className="v-align-center">
+
+const Offer = ({offer, prices, onClick}) => {
+  const price = prices[offer.token.symbol][offer.currency];
+  const calcPrice = offer.marketType === "0" ? price * ((100 + (parseFloat(offer.margin))) / 100) : price * parseFloat(offer.margin) / 100;
```
jrainville: I feel like this formula is gonna get used a few times to calculate the price with the margins, We might want to move it to a utils.

src/js/pages/Profile/index.scss
```diff
@@ -8,4 +8,8 @@
     overflow: hidden;
     border-radius: 10px;
   }
+
+  .btn {
+    padding: 11px;
```
jrainville: Bootstrap had `.p-x` (where x = 1, 2, 3, etc) if you prefer. Disregard if you have a lot of buttons that would require copying the class a lot

PR #174, opened 2019-04-19
fix: hide dispute btn when offer is created and allow contacting seller

richard-ramos:


src/js/pages/Escrow/components/CardEscrowBuyer.jsx
```diff
@@ -84,17 +84,21 @@ Funded.propTypes = {
...
-const PreFund = () => (
+const PreFund = (statusContactCode) => (
   <React.Fragment>
     <span className="bg-dark text-white p-3 rounded-circle">
       <img src={one} alt="one" />
     </span>
     <p className="h2 mt-4">Waiting for the seller to fund an escrow</p>
     <p>Notify the seller about the trade using Status encrypted p2p chat</p>
-    <Button color="primary" className="btn-lg mt-3" onClick={() => {}}>Open chat</Button>
+    <a href={"https://get.status.im/user/" + statusContactCode} className="btn btn-primary btn-lg" role="button" aria-disabled="true">Open chat</a>
```
jrainville: We probably want to open it in a new tab with `target="_blank"`

PR #175, opened 2019-04-19
feat: use correct icons and adds a seller/buyer profile action button

richard-ramos:


src/js/pages/Escrow/index.jsx
```diff
@@ -150,7 +151,9 @@ class Escrow extends Component {
             <Offer offer={offer} prices={prices}/>
           </Col>
         </Row>
-        <OpenChat statusContactCode={offer.user.statusContactCode} />
+        <OpenChat statusContactCode={offer.user.statusContactCode} withBuyer={!isBuyer} />
+        <Profile withBuyer={!isBuyer} address={isBuyer ? escrow.offer.owner : escrow.buyer} />
+        <hr />
```
jrainville: It's better to use borders if possible

richard-ramos: I think in this case it's okay because in the design there's a margin-top|bottom between the HR and the elements.

PR #180, opened 2019-04-19
fix: set trade status colors

richard-ramos:


src/js/pages/MyProfile/components/Trades.jsx
```diff
@@ -4,9 +4,31 @@ import {Card, Row, Col} from 'reactstrap';
...
+    case tradeStates.canceled:
+      return {text: 'Canceled', color: '#939BA1'};
+    case tradeStates.expired:
+      return {text: 'Expired', color: '#939BA1'};
+    case tradeStates.arbitration_open:
+      return {text: 'Arbitration', color: '#FF2D55'};
+    case tradeStates.arbitration_closed:
+      return {text: 'Resolved', color: '#4360DF'};
+    default:
+      return {text: tradeState, color: '#939BA1'};
```
jrainville: Would be better to use a class name instead of putting the color codes here. You can see in variable-overrides.scss, all the colors are already there: https://github.com/status-im/status-teller-network/blob/master/src/css/variable-overrides.scss#L39-L49

Most of them already have classes too:
- blue: `text-primary`
- red : `text-danger`
- green: `text-success`

It's in the same file: https://github.com/status-im/status-teller-network/blob/master/src/css/variable-overrides.scss#L72-L80

src/js/pages/MyProfile/index.scss
```diff
@@ -22,4 +23,20 @@
...
+.status-gray {
+  background-color: $dark;
+}
+
+.status-blue {
+  background-color: $blue;
+}
+
+.status-red {
+  background-color: $red;
```
jrainville: Sorry to make you make you do stuff then erase, but bootstrap already has classes for that. So instead of creating `status-color` classes, you can just use the `.bg-color` classes: https://getbootstrap.com/docs/4.3/utilities/colors/#background-color

Again,
red = danger
green = success
blue = primary
dark = dark (lucky) (could be secondary depending on your choice though)

src/js/pages/MyProfile/index.scss
```diff
@@ -1,3 +1,4 @@
+@import "../../../css/variable-overrides";
```
jrainville: No longer need this

richard-ramos: :'(

jrainville: I will never approve this PR!!! 😈

richard-ramos: LOL!

PR #185, opened 2019-04-20
fix: change market button for radio buttons and made the location screen match the design

richard-ramos:


iurimatias:
@richard-ramos travis failing

richard-ramos:
Fixed, I think. It's the tests that are failing. I updated them with `yarn run test:unit -u`

PR #187, opened 2019-04-23
feat: update arbitration contract in order to require a license for being an arbitrator.

richard-ramos:


contracts/teller-network/License.sol
```diff
@@ -73,66 +63,13 @@ contract License is Ownable, ApproveAndCallFallBack {
...
         idxLicenseOwners[msg.sender] = idx;
 
         emit Bought(_owner, token.allowance(_owner, address(this)));
     }
 
-    /**
-    * @dev Release a license and retrieve funds
-    * @notice Only the owner of a license can perform the operation after the release delay time has passed.
-    */
-    function release() public {
```
jrainville: This was just old code we never used, right?

contracts/teller-network/Escrow.sol
```diff
@@ -436,16 +436,15 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable {
      * @notice Set arbitration result in favour of the buyer or seller and transfer funds accordingly
      * @param _escrowId Id of the escrow
      * @param _releaseFunds Release funds to buyer or cancel escrow
+     * @param _arbitrator Arbitrator address
      */
-    function setArbitrationResult(uint _escrowId, bool _releaseFunds) external {
+    function setArbitrationResult(uint _escrowId, bool _releaseFunds, address _arbitrator) external {
         assert(msg.sender == address(arbitration)); // Only arbitration contract can invoke this
```
jrainville: ~~Why do we need to pass the `_arbitrator` if we check here that the sender is `address(arbitration)`? ~~
Nvm, jus saw, `abitration` is the address of the contract.

jrainville: Also, this PR is to make it so that anyone can be an arbiter as long as it's not the buyer or seller? Is that so?

PR #191, opened 2019-04-30
feat: open dispute screen

richard-ramos:


src/js/pages/MyProfile/components/Trades.jsx
```diff
@@ -50,7 +60,7 @@ class Trades extends Component {
                 {isBuyer ? 'Buy' : 'Sell' } {formatBalance(trade.tokenAmount)} {trade.token.symbol}
               </Col>
               <Col className="align-self-center text-center text-success" xs="4">
-                <span className={"p-1 text-uppercase d-inline text-white rounded-sm " + tradeStyle.className}>{tradeStyle.text}</span>
+                <span className={"p-1 text-uppercase d-inline text-white rounded-sm " + tradeStyle.className}><small>{tradeStyle.text}</small></span>
```
jrainville: I added the class `.text-small` or something like that. It puts the text in 13px. Is that what your are searching for with the `<small>`?

src/js/pages/OpenDispute/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+  goToProfile = () => {
+    this.props.history.push('/profile');
+  }
+
+
+  render(){
+    const {escrow, openDispute, loading, receipt} = this.props;
+
+    if(!escrow) return <Loading page={true} />;
+    if(loading) return <Loading mining={true} />;
```
jrainville: FYI, you can just do `<Loading page />`

src/js/pages/OpenDispute/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+        <p className="text-muted">Follow the progress of your dispute in the profile.</p>
+        <p>
+          <Button color="primary" onClick={this.goToProfile}>Okay</Button>
+        </p>
+    </div>
+    );
+
+    const onClickDialog = () => {
+      openDispute(escrow.escrowId);
+    };
```
jrainville: This function can be put outside of `render` so that is not defined every render

src/js/pages/OpenDispute/index.jsx
```diff
@@ -0,0 +1,120 @@
...
+            name="disputeDetails"
+            id="disputeDetails"
+            rows="7"
+            placeholder="What has happened?"
+            className="form-control mb-2"
+            value={this.state.value}
+            onChange={this.handleChange}
+            validations={[]}
+          />
+          <p className="text-muted">The process of resolving your dispute could take up on 5 days. You will be sharing you chat logs with an arbiter.</p>
```
jrainville:
```suggestion
          <p className="text-muted">The process of resolving your dispute could take up on 5 days. You will be sharing your chat logs with an arbiter.</p>
```

src/js/pages/OpenDispute/index.jsx
richard-ramos: Cool! didn't know that tip!

PR #198, opened 2019-05-06
Fix Map in Status + shared chains

jrainville:
- Fix the infite map reload on Status by displaying the Map + the error
 - I couldn't see what the actual error is because Google doesn't like localhost network connections, but next time we are on a deployed env, we'll see it
- Add shared chains.json for testnets
- Add embark-solc because gotta go fast

### DIFF
--- a/embarkConfig/contracts.js
+++ b/embarkConfig/contracts.js
@@ -139,11 +139,12 @@ module.exports = {
   // merges with the settings in default
   // used with "embark run testnet"
   testnet: {
+    tracking: 'shared.chains.json',
     deployment: {
       accounts: [
         {
           mnemonic: secret.mnemonic,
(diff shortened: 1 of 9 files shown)


embarkConfig/contracts.js
```diff
@@ -139,11 +139,12 @@ module.exports = {
   // merges with the settings in default
   // used with "embark run testnet"
   testnet: {
+    tracking: 'shared.chains.json',
     deployment: {
       accounts: [
         {
           mnemonic: secret.mnemonic,
-          hdpath: "m/44'/1'/0'/0/",
+          hdpath: secret.hdpath || "m/44'/60'/0'/0/",
```
jrainville: Because Iuri has a weird HDPath

PR #208, opened 2019-05-09
feat: use contact code as seed for the identicons

richard-ramos:
this will guarantee that once the "Open chat" buttons are pressed, the identicon presented for the user will match the one on the escrow

### DIFF
--- a/src/js/components/UserInformation/Address.scss
+++ b/src/js/components/UserInformation/Address.scss
@@ -1,0 +1,3 @@
+.addr {
+  word-wrap: break-word;
+}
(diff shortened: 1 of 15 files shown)


src/js/components/UserInformation/Address.scss
```diff
@@ -0,0 +1,3 @@
+.addr {
+  word-wrap: break-word;
```
jrainville: Boot strap has a class for that: https://embark.status.im/docs/contracts_configuration.html#Providing-ABIs

PR #219, opened 2019-05-12
add generic token; fix license price issue when error occurs

iurimatias:


src/js/features/license/selectors.js
```diff
@@ -5,4 +5,9 @@ export const error = state => state.license.error;
 export const txHash = state => state.license.txHash;
 export const licenseOwners = state => state.license.licenseOwners;
 export const licenseOwnersError = state => state.license.licenseOwnersError;
-export const getLicensePrice = state => parseInt(state.license.price, 10);
+export const getLicensePrice = state => {
+  if (state.license.price === Number.MAX_SAFE_INTEGER) {
+    return "price_error";
+  }
```
jrainville: This is supposed to return a number, so we probably have bad error handling in terms of when we get the prices.

embarkConfig/contracts.js
```diff
@@ -113,7 +113,98 @@ module.exports = {
...
+      // PRG: { instanceOf: "StandardToken" },
+      // DPY: { instanceOf: "StandardToken" },
+      // CDT: { instanceOf: "StandardToken" },
+      // TNT: { instanceOf: "StandardToken" },
+      // DRT: { instanceOf: "StandardToken" },
+      // SPANK: { instanceOf: "StandardToken" },
+      // BRLN: { instanceOf: "StandardToken" },
+      // USDC: { instanceOf: "StandardToken" },
+      // LPT: { instanceOf: "StandardToken" },
+      // ST: { instanceOf: "StandardToken" }
```
jrainville: Can we put this in another file or function. We'll probably need to reuse it too (or maybe put in default)

PR #220, opened 2019-05-12
Misc refactors

iurimatias:


src/js/components/EditContact/validators.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+};
+
+export const required = (value) => {
+  if (!value.toString().trim().length) {
+    return <NamespacesConsumer>
+      {t => <FormFeedback className="d-block">{t('validators.required')}</FormFeedback>}
+    </NamespacesConsumer>;
+  }
+};
```
jrainville: This function didn't need to be moved

src/js/components/Map/index.jsx
```diff
@@ -7,27 +7,30 @@ import {withNamespaces} from 'react-i18next';
...
 export class Map extends Component {
   constructor(props) {
+    const coords = props.coords || {};
     super(props);
+
     this.state = {
       activeMarkers: {},
-      center: {lat: props.coords.latitude, lng: props.coords.longitude}
+      center: {lat: coords.latitude, lng: coords.longitude},
+      error: props.error
```
jrainville: You never use this state property

PR #227, opened 2019-05-13
fix: sidebar css for displaying checkboxes no matter the screen size

richard-ramos:


src/js/pages/OffersList/components/SorterFilter.scss
```diff
@@ -40,7 +40,7 @@ $menu-width: 85%;
 
   &.open {
     width: $menu-width;
-    padding: 24px $filter-menu-padding-x;
+    padding: 24px;
```
jrainville: Does this make the padding too big for small screens?

richard-ramos: In small screens it looked okay, however at bigger screen sizes this happens:
Screenshot

jrainville: This screen capture is before or after the fix?

richard-ramos: before the fix

jrainville: Ah ok, I'm curious now with the fix if on small screens the padding is too big

Issue #206, opened 2019-05-09
MetadataStore#addOrUpdateUser is open


3esmit:
### Bug Report
Function allows access by anyone to anyone.

#### Summary
MetadataStore.sol function addOrUpdateUser(address _user, bytes _statusContactCode, string _location, string _username) is bad. _user parameter is open, so anyone can change anyone information. 

#### Suggested Fix
Functions need to consider msg.sender for defining the address of _user:

```
function addOrUpdateUser(bytes _statusContactCode, string _location, string _username) external {
address _user = msg.sender;
(...)
}
```

I see that this "open" is a feature to allow new buyers having their data being set by seller. To keep this capability, but safe, require a ethereum signed message containing the contract address and the parameters. The recovered address of the signature is set as _user.

```
function addOrUpdateUser(bytes _statusContactCode, string _location, string _username, bytes _signature) public {
address _user = recoverAddress(keccak256(address(this), _statusContactCode, _location, _username), _signature);
(...)
}
```

arsenyjin:
https://github.com/status-im/status-teller-network/pull/228

arsenyjin:
I guess we can close the issue because of the  https://github.com/status-im/status-teller-network/pull/314

PR #228, opened 2019-05-14
Change MetadataStore#addOrUpdateUser

arsenyjin:
Added signature to addOrUpdateUser as mentioned in https://github.com/status-im/status-teller-network/issues/206


3esmit:
This solves the problem of the issue, so the metadata is not wide open anymore. However we should rethink about the reuse of signatures. Perhaps we can consider adding TTL, nonce (or orderid, if we keep Escrow and Metadata related).

iurimatias:
the tests currently fail

arsenyjin:
I probably have to modify tests then, as I've changed the way several functions work

arsenyjin:
Closed the PR as signature mechanics probably need more modifications

PR #233, opened 2019-05-14
fix: change escrow status

richard-ramos:


src/js/features/escrow/reducer.js
```diff
@@ -51,6 +51,7 @@ function reducer(state = DEFAULT_STATE, action) {
         fundEscrowStatus: States.failed
       };
     case FUND_ESCROW_SUCCEEDED:
+      state.escrow.status = escrowStatus.FUNDED;
```
jrainville: I don't think you're supposed to modify the state directly like this.
I think you should do something like that instead:
```
return {
        ...state,
        escrow: {
          ...state.escrow,
          status: escrowStatus.FUNDED
        }
        fundEscrowStatus: States.success
      };
```

jrainville: And same thing for the others obviously

richard-ramos: > We can't modify the state directly.
Also, should we add the failing states too?
I'm asking because I'm not super sure what this change actually gives us, since we already have the fundEscrowStatus. Why putting it twice?

This is for displaying the buttons correctly in the escrow page. Those buttons will hide/show depending on the escrow's current status.

jrainville: Aaah ok, I get it now. I had not realized that the Escrow status was different than tx status (funded != success)

PR #236, opened 2019-05-15
feat: arbitrator selector screen

richard-ramos:


contracts/teller-network/MetadataStore.sol
```diff
@@ -179,7 +189,8 @@ contract MetadataStore is Ownable {
         string memory _username,
         PaymenMethods[] memory _paymentMethods,
         MarketType _marketType,
-        uint8 _margin
+        uint8 _margin,
+        address _arbitrator
     ) public {
         require(userWhitelist[msg.sender], "User does not exist");
         require(offerWhitelist[msg.sender][_offerId], "Offer does not exist");
```
jrainville: We should add a require to make sure the arbitrator really is an arbitrator

PR #241, opened 2019-05-16
fix: show dispute status in escrow page

richard-ramos:


src/js/pages/Escrow/components/CardEscrowBuyer.jsx
```diff
@@ -107,13 +107,33 @@ PreFund.propTypes = {
   statusContactCode: PropTypes.string
 };
 
+const Dispute = () => (
+  <React.Fragment>
```
jrainville: Just a tip: you can do `<Fragment>` now. (just need to import it above like so: `import React, {Component, Fragment} from 'react';`

src/js/pages/Escrow/components/CardEscrowBuyer.jsx
```diff
@@ -107,13 +107,33 @@ PreFund.propTypes = {
...
+
 class CardEscrowBuyer extends Component {
   render(){
-    const {trade, payStatus, payAction, rateTransaction} = this.props;
+    const {trade, payStatus, payAction, rateTransaction, arbitrationDetails} = this.props;
 
     const showLoading = payStatus === States.pending;
     const showWaiting = payStatus === States.success || trade.status === escrow.helpers.tradeStates.released;
 
+    if(arbitrationDetails && (arbitrationDetails.open || arbitrationDetails.result !== "0")){
```
jrainville: Is there a chance that the result could be an integer instead of a string?

src/js/pages/Escrow/components/CardEscrowSeller.jsx
```diff
@@ -121,6 +121,14 @@ Start.propTypes = {
   onClick: PropTypes.func
 };
 
+const Dispute = () => (
```
jrainville: We use this twice. We could move it to its own file

src/js/pages/Escrow/components/CardEscrowBuyer.jsx
richard-ramos: I'll add a `.toString()`  just in case there's the possibility of it being a number

PR #243, opened 2019-05-16
fix: buyer cannot buy from themselves

richard-ramos:


src/js/pages/Profile/components/Offer.jsx
```diff
@@ -6,21 +6,26 @@ import {TokenImages} from '../../../utils/images';
...
 
-const Offer = ({offer, prices, onClick}) => (
+const Offer = ({offer, prices, onClick, disabled}) => (
   <Row className="border py-2 mx-0 my-2 rounded">
     <Col xs="4" className="v-align-center">
       <p className="font-weight-bold"><img src={TokenImages[`${offer.token.symbol}.png`] || TokenImages[`generic.png`]} alt="asset icon" className="mr-2"/>{offer.token.symbol}</p>
     </Col>
     <Col xs="8" className="v-align-center text-right">
-      <Button color="primary p-2" onClick={onClick}>Buy for {truncateTwo(calculateEscrowPrice(offer, prices))} {offer.currency}</Button>
+      <Button color={disabled ? "secondary p-2" : "primary p-2" } disabled={disabled} onClick={onClick}>Buy for {truncateTwo(calculateEscrowPrice(offer, prices))} {offer.currency}</Button>
```
jrainville: You cannot put `p-2` in color lol.

src/js/pages/Profile/index.jsx
```diff
@@ -11,13 +12,17 @@ import Offer from './components/Offer';
...
 import newBuy from "../../features/newBuy";
+import network from "../../features/network";
 
 import './index.scss';
 import Loading from "../../components/Loading";
 
 class Profile extends Component {
   componentDidMount() {
     this.props.load(this.props.match.params.address);
+    if(web3.utils.toChecksumAddress(this.props.match.params.address) === this.props.address){
```
jrainville: You should put this`if` above the load, and return so that we don't load for no reason

src/js/pages/Profile/components/Offer.jsx
richard-ramos: oops!  I did not read the name of the attribute!

PR #244, opened 2019-05-16
fix: make the sidebar filter scrollable on smaller devices

richard-ramos:


src/js/pages/OffersList/components/SorterFilter.scss
```diff
@@ -55,6 +55,13 @@ $menu-width: 85%;
     position: fixed;
     max-width: 350px;
     height: 100%;
+    padding-bottom: 50px;
+  }
+
+  @media (max-width: 767px) {
+    .filter-menu-content {
+      overflow-y: scroll;
```
jrainville: Can't we just always put `overflow-y: auto;`?

richard-ramos: yup! :D

PR #245, opened 2019-05-16
Improve state management of escrows

jrainville:
This is a pretty big one in the fact that it touches the whole state management of the escrows.

The goal: make the flow more reactive and more informative.

With this, when a Tx is sent, the user gets a link to etherscan with the tx hash. He can also go back to the profile and see the actual state of the TX and go back to it and when it's done, all the states are correctly updated.

Includes a couple of other fixes and improvement, but here is what the new states in the list look like:
image

### DIFF
--- a/src/js/pages/Escrow/components/CardEscrowSeller.jsx
+++ b/src/js/pages/Escrow/components/CardEscrowSeller.jsx
@@ -68,8 +68,8 @@ class PreFund extends Component {
       return <Loading page={true}/>; // Wait for trade to be populated
     }
 
-    const enoughBalance = toBN(trade.token.balance ? toTokenDecimals(trade.token.balance, trade.token.decimals) : 0).gte(toBN(trade.tradeAmount)) &&
-                          toBN(toTokenDecimals(tokens.SNT.balance, 18)).gte(toBN(fee));
+    const enoughBalance = toBN(trade.token.balance ? toTokenDecimals(trade.token.balance || 0, trade.token.decimals) : 0).gte(toBN(trade.tradeAmount)) &&
+                          toBN(toTokenDecimals(tokens.SNT.balance || 0, 18)).gte(toBN(fee));
     return <Fragment>
       <span className="bg-dark text-white p-3 rounded-circle">
         <img src={two} alt="two" />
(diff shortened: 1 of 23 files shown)


src/js/pages/Escrow/components/CardEscrowSeller.jsx
```diff
@@ -80,8 +68,8 @@ class PreFund extends Component {
       return <Loading page={true}/>; // Wait for trade to be populated
     }
 
-    const enoughBalance = toBN(trade.token.balance ? toTokenDecimals(trade.token.balance, trade.token.decimals) : 0).gte(toBN(trade.tradeAmount)) &&
-                          toBN(toTokenDecimals(tokens.SNT.balance, 18)).gte(toBN(fee));
+    const enoughBalance = toBN(trade.token.balance ? toTokenDecimals(trade.token.balance || 0, trade.token.decimals) : 0).gte(toBN(trade.tradeAmount)) &&
```
richard-ramos: is it possible for the balance to return undefined or null?

src/js/ui/TxHash/index.js
```diff
@@ -0,0 +1,34 @@
...
+import {getNetwork} from "../../features/network/selectors";
+import {connect} from "react-redux";
+
+const etherScanUrls = {
+  1: 'https://etherscan.io/tx/',
+  3: 'https://ropsten.etherscan.io/tx/',
+  4: 'https://rinkeby.etherscan.io/tx/',
+  5: 'https://goerli.etherscan.io/tx/',
+  42: 'https://kovan.etherscan.io/tx/',
+  401697: 'https://tobalaba.etherscan.com/tx/'
```
richard-ramos: TIL about tobalaba!

src/js/ui/TxHash/index.js
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const TxHash = ({value, network}) => {
+  if (!network || (!network.id && network.id !== 0) || !etherScanUrls[network.id]) {
+    return value;
+  }
+
+  return (<a href={etherScanUrls[network.id] + value} target="_blank" rel="noopener noreferrer">{value}</a>);
```
richard-ramos: What do you think of adding a confirmation that the user is leaving the teller network dapp? Status does not support atm opening links in new tabs.

src/js/pages/Escrow/components/CardEscrowSeller.jsx
jrainville: Yeah it happened to me sometimes and it would cause a crash. So that makes sure it doesn't crash while the balance is loading.

src/js/ui/TxHash/index.js
jrainville: Same haha

src/js/ui/TxHash/index.js
jrainville: Yeah that's an idea. It makes me think, maybe it supports opening in a new window? Maybe Status is stupid and it doesn't have tabs, but it has windows?
I'll give it a try

jrainville:
Trades not appearing fixed here: https://github.com/status-im/status-teller-network/pull/245/commits/b6cc6217b6945866f72bce4e7a9a0ecb287e7e71

src/js/ui/TxHash/index.js
jrainville: Mind if I do that in another PR? This one is already complex enough without adding another layer. (gonna add it to Pivotal)

richard-ramos: np!

PR #247, opened 2019-05-17
feat: sellers cannot cancel an escrow until after the expiration time

richard-ramos:


src/js/pages/Escrow/index.scss
```diff
@@ -4,4 +4,8 @@
   p.balanceAlert {
     color: #FF2D55;
   }
-}
\ No newline at end of file
+
+  .disabled {
+    opacity: .4;
```
jrainville: Using `.text-muted` should do the same effect

richard-ramos: Ah! I'm using `.disabled` to to set the image of the button as semitransparent too

jrainville: Ah ok, I didn't notice. It's fine then!

PR #248, opened 2019-05-17
fix: sellers shouldnt be their own arbitrator

richard-ramos:


contracts/teller-network/License.sol
```diff
@@ -61,13 +62,15 @@ contract License is Ownable, ApproveAndCallFallBack {
...
-        licenseDetails[_owner].creationTime = block.timestamp;
+        licenseDetails[_owner] = LicenseDetails({
+            price: price,
+            creationTime: block.timestamp
+        });
 
-        uint idx = licenseOwners.push(msg.sender);
-        idxLicenseOwners[msg.sender] = idx;
+        uint idx = licenseOwners.push(_owner);
+        idxLicenseOwners[_owner] = idx;
```
jrainville: Does that mean that you can buy a license for someone else?

richard-ramos: no, because `buyFrom(_from)` is a private function. It can only be called in two places:
1. by using `buy()` which will send the `msg.sender`,  
2. it is called by the `receiveApproval` function that sends the `msg.sender` in the `_from` parameter. (this is to approve and execute the function immediatly without requiring two transactions).

jrainville: Ok, thanks for the explanation

PR #251, opened 2019-05-17
Set max as sellers balance

jrainville:
Also change the color of the offer when it's your own offer


jrainville:
Gonna merge this now and do the handling of balances in another PR

PR #252, opened 2019-05-20
fix: add logic for "make a decision" button in arbitration

richard-ramos:


src/js/pages/Arbitration/index.jsx
```diff
@@ -36,20 +40,58 @@ class Arbitration extends Component {
...
   }
 
+  state = {
+    displayUsers: false,
+    selectedUser: null,
+    displayDialog: false
+  }
+
   componentDidMount() {
   }
```
jrainville: We can remove this

src/js/pages/Arbitration/index.jsx
```diff
@@ -36,20 +40,58 @@ class Arbitration extends Component {
...
 
+  openSolveDisputeDialog = () => {
+    this.setState({displayUsers: true});
```
jrainville: Shouldn't this be `displayDialog`? Or it's just the name of the function that's confusing?

src/js/pages/Arbitration/index.jsx
```diff
@@ -36,20 +40,58 @@ class Arbitration extends Component {
...
+    const {displayUsers, selectedUser} = this.state;
 
     if(!escrow){
       return <Loading/>;
     }
 
+    if(loading) return <Loading mining={true} />;
+
     if(escrow.buyer === address || escrow.seller === address) return <ErrorInformation message="You cannot arbitrate your own disputes"/>;
+    if(escrow.arbitrator !== address) return <ErrorInformation message="You are not the arbitrator of this dispute"/>;
```
jrainville: It's better to put the errors (both of them, about the loadings, to avoid infinite loadings

src/js/pages/Arbitration/index.jsx
```diff
@@ -65,13 +107,37 @@ class Arbitration extends Component {
...
+                    <Identicon seed={escrow.buyerInfo.statusContactCode} className="rounded-circle border mr-2" scale={5}/>
+                    {escrow.buyerInfo.username}
+                  </CheckButton>
+                  <CheckButton active={selectedUser === escrow.seller} size="l" onClick={this.selectUser(escrow.seller)}>
+                    <Identicon seed={escrow.sellerInfo.statusContactCode} className="rounded-circle border mr-2" scale={5}/>
+                    {escrow.sellerInfo.username}
+                  </CheckButton>
+                </ButtonGroup>
+                <p className="text-center">
+                  <Button color="primary" onClick={this.displayDialog(true)} disabled={selectedUser === null}>Resolve dispute</Button>
```
jrainville: We should probably add a "cancel" button too

src/js/pages/Arbitration/index.scss
```diff
@@ -13,6 +13,10 @@
   padding: 3px 10px;
 }
 
+.arbitrationDialog button svg {
+  margin-top: 10px !important;
```
jrainville: Can you use `mt-2` instead? (you can change the number too)

src/js/pages/Arbitration/index.jsx
richard-ramos: there are two display functions here. One to select the user that will win the dispute, and another to confirm the selection.

src/js/pages/Arbitration/index.scss
richard-ramos: I tried and the checkbox refused to move :(

src/js/pages/Arbitration/index.jsx
jrainville: Above the other loading too 🤷‍♂

richard-ramos: ah but does it make sense? the other loading happens only while `escrow == null`  and these errors depend on the escrow not being null.

jrainville: Oops, you are totally right, sorry

PR #260, opened 2019-05-22
fix: dispute state after arbitration

richard-ramos:


src/js/features/arbitration/reducer.js
```diff
@@ -83,13 +83,26 @@ function reducer(state = DEFAULT_STATE, action) {
...
+      const currentEscrow = escrows.find(x => x.escrowId === action.escrowId);
+      if(currentEscrow) {
+        currentEscrow.arbitration.open = false;
+        currentEscrow.arbitration.result = action.result;
+      }
+
       return {
         ...state, ...{
           escrows,
+          arbitration,
```
jrainville: To not modify the state directly:
```suggestion
          arbitration: {
            ...state.arbitration,
            arbitration: {
              arbitration: {
                open: false,
                result:  action.result
              }
            }
          },
```
You can then get rid of the code above

src/js/features/arbitration/reducer.js
```diff
@@ -83,13 +83,26 @@ function reducer(state = DEFAULT_STATE, action) {
...
+    {
+      const arbitration = state.arbitration;
+      arbitration.arbitration.open = false;
+      arbitration.arbitration.result = action.result;
+
+      const currentEscrow = escrows.find(x => x.escrowId === action.escrowId);
+      if(currentEscrow) {
+        currentEscrow.arbitration.open = false;
+        currentEscrow.arbitration.result = action.result;
+      }
```
jrainville: This modifies the state directly and also, I just noticed that we have escrows inside the arbitration reducer. Does it make sense?
Do we duplicate the escrows in two reducers or in this one it's only about the arbitrations in escrows?

src/js/features/arbitration/reducer.js
richard-ramos: ahh! I always forget this. Sorry!

src/js/features/arbitration/reducer.js
richard-ramos: It's only arbitrations in escrows!

PR #269, opened 2019-05-24
Remove useless blockchain calls and use the state instead

jrainville:
We fetched the buyer and seller info from the chain for arbitrations when it was useless. That's the responsibility of the metadata state, so I moved that logic in the Dispute component and we only load the profiles if they weren't already loaded.


jrainville:
Oops, pushed my other commit in the same branch. Oh well, it's just styling. 
Could you take another quick look (only at https://github.com/status-im/status-teller-network/pull/269/commits/c26b89f280993c873c96b011849264c8f8e28cc4), @richard-ramos ?

PR #273, opened 2019-05-24
Show who won and lost arbitrations

jrainville:
Arbitrator's page:
arbitration
Loser's profile:
profile
Winner's escrow:
escrow

### DIFF
--- a/src/js/pages/Arbitration/index.jsx
+++ b/src/js/pages/Arbitration/index.jsx
@@ -151,6 +151,12 @@ class Arbitration extends Component {
     return (
       <div className="escrow">
         <h2>Dispute Details <span className={"arbitrationStatus " + status}>{status}</span></h2>
+
+        {escrow.arbitration.result.toString() !== "0" && <Fragment>
+          <h3>Dispute resolved</h3>
+          <p><span className="font-weight-bold">Winner:</span> {escrow.arbitration.result.toString() === ARBITRATION_SOLVED_BUYER ? 'Buyer' : 'Seller'}</p>
+        </Fragment>}
+
         <p className="arbitrationMotive mt-3 mb-0">{escrow.arbitration.motive}</p>
         <span className="triangle"><img src={bubbleTriangle} alt="buble-triangle"/></span>
         <TradeParticipant address={escrow.arbitration.openBy}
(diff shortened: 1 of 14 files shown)


src/js/pages/Arbitration/index.jsx
```diff
@@ -151,6 +151,12 @@ class Arbitration extends Component {
     return (
       <div className="escrow">
         <h2>Dispute Details <span className={"arbitrationStatus " + status}>{status}</span></h2>
+
+        {escrow.arbitration.result.toString() !== "0" && <Fragment>
```
richard-ramos: What do you think of using the constant ARBITRATION_UNSOLVED instead of "0"?  I probably have to change that in a couple of places too

jrainville: good idea. I'll do the changes

src/js/pages/Escrow/components/CardEscrowSeller.jsx
```diff
@@ -167,43 +167,44 @@ class CardEscrowSeller extends Component {
...
+    if (fundStatus === States.success) step = 4;
+    if (releaseStatus === States.pending || (trade.mining && (trade.status === escrow.helpers.tradeStates.funded || trade.status === escrow.helpers.tradeStates.paid))) step = 5;
+    if (releaseStatus === States.success || trade.status === escrow.helpers.tradeStates.released) step = 6;
+    if (arbitrationDetails && arbitrationDetails.open && arbitrationDetails.result.toString() === "0") step = 10;
+    if (arbitrationDetails && arbitrationDetails.result.toString() !== "0") step = 11;
 
     let component;
-    switch(step){
+    switch (step) {
+      case 10:
```
richard-ramos: Not for this PR, but I'm thinking that we need to move away from this switch to something more robust

jrainville: Agreed.
I tried changing it, but it's a lot more massive than I anticipated, so I reverted. 
We need a PR just for that, because there's a lot of possibility of breaking it

src/js/pages/Escrow/components/ResolvedDispute.jsx
```diff
@@ -0,0 +1,18 @@
+import React, {Fragment} from 'react';
+import PropTypes from 'prop-types';
+import {faTimes, faCheck} from "@fortawesome/free-solid-svg-icons";
+import RoundedIcon from "../../../ui/RoundedIcon";
+
+const ResolvedDispute = ({winner}) => (
+  <Fragment>
+    <RoundedIcon icon={winner ? faCheck : faTimes} bgColor={winner ? 'green' : 'red'}/>
+    <h2 className="mt-4">Dispute Resolved</h2>
+    <p className="m-0">The dispute was resolved and you are the {winner ? 'winner' : 'loser'}</p>
```
richard-ramos: Maybe changing this to "The disputes was resolved in your favor", "The dispute was solved in favor of the (seller|buyer)" ?

PR #274, opened 2019-05-27
feat(margin): use a slider for the margin selector

jrainville:
image

### DIFF
--- a/src/js/wizards/Sell/6_Margin/components/MarginSelectorForm.jsx
+++ b/src/js/wizards/Sell/6_Margin/components/MarginSelectorForm.jsx
@@ -1,17 +1,32 @@
 /* global web3 */
 import React, {Component, Fragment} from 'react';
 import PropTypes from 'prop-types';
-import {FormGroup, InputGroup, InputGroupAddon, ButtonGroup, InputGroupText} from 'reactstrap';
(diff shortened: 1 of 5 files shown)


src/js/wizards/Sell/6_Margin/components/MarginSelectorForm.jsx
```diff
@@ -1,53 +1,69 @@
...
+import './MarginSelectorForm.scss';
 
 const ABOVE = 0;
 const BELOW = 1;
 
 class MarginSelectorForm extends Component {
+  onMarginChange = (value) => {
+    if (value < 0) {
+      if (this.props.marketType === ABOVE) {
+        this.props.marketTypeChange(BELOW);
```
richard-ramos: Maybe this could be refactored to not having to use a market type but read if the value is greater or minor than 0 to know the type of market.
what do you think?

jrainville: Yeah, that's pretty much what I did in the UI part of it.
Basically, the Container and actions are still using the market type, but the presentation layer works with -100 to 100 and converts to market type.
I agree that we could remove the market type, but I didn't want to have to modify the contract just now.
I'll add a task for it though.

PR #280, opened 2019-05-28
feat: cancel arbitration UI + feature

richard-ramos:


src/js/features/arbitration/reducer.js
```diff
@@ -59,12 +64,14 @@ function reducer(state = DEFAULT_STATE, action) {
           loading: false
         }
       };
+    case CANCEL_DISPUTE_SUCCEEDED:
```
jrainville: On success, don't you have to change the escrow's state?

PR #282, opened 2019-05-28
Watch escrow events to make workflow reactive

jrainville:
Add event listeners to the escrows so taht if a user does an action, the other party will get the updated state once the tx is done.
Makes sure to only subscribe to events once (per escrow).
This also includes utils functions and bootstraping for future events we might want to use.


richard-ramos:
Very good work!

PR #283, opened 2019-05-28
fix: allow buyers to cancel an escrow

richard-ramos:


contracts/teller-network/Escrow.sol
```diff
@@ -352,7 +354,7 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable {
         require(_escrowId < transactions.length, INVALID_ESCROW_ID);
```
jrainville: I see this missed the check so that only the seller can do it. Or maybe we change the cancel function so that it transfers to the seller all the time and not the `msg.sender`

richard-ramos: good catch, dude!

jrainville: I'm an auditor now 🕵

PR #294, opened 2019-06-03
feat: add gasless functionality to escrow contract

richard-ramos:
1. Adds procedure to run a tabookey relayer locally
2. Adds gasless functionality for contract to only apply when the user has < 500000*gasPrice in balance, and only ETH offers with some throttling restrictions
3. Ropsten network section (no contract config yet)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
 2. `yarn start` (starts create-react-app pipeline)
 
 ## Running Tabookey Gas Relayer (no docker. For docker, see their README)
+
 1. Clone v3.0.0 of tabookey-gasless
 ```
 git clone https://github.com/tabookey/tabookey-gasless
@@ -20,19 +21,20 @@ git checkout 2316c7422d50ac0242f8442f6dc98d0c85512c13
 npm install
 npm test
 ```
-2. Run the server:
+
+2. Run the server. (Embark must be running, and it should have deployed the RelayHub contract)
 ```
-./build/server/bin/RelayHttpServer -RelayHubAddress 0x15d6e765aA9DaeBDdEB14141a35187FC850f87A3 -Workdir ./build/server --EthereumNodeUrl http://localhost:8555
+./build/server/bin/RelayHttpServer -RelayHubAddress RELAY_HUB_CONTRACT_ADDRESS_HERE -Workdir ./build/server --EthereumNodeUrl http://localhost:8555
 ```
 Replace the RelayHub contract address for the correct address.
 3. Browse http://localhost:8090/getaddr. Copy the address
 4. Stake ether for that address. You can execute this in the embark console:
 ```
-RelayHub.methods.stake("0x4e89c5FDf14877DA9757ea4c0FdC901023E5a868", 30).send({value: web3.utils.toWei("1", "ether"), gas:800000}) 
+RelayHub.methods.stake(RELAYER_ADDRESS_GOES_HERE, 30).send({value: web3.utils.toWei("1", "ether"), gas:800000}) 
 ```
 4. Send some ether to the relayer
 ```
-web3.eth.sendTransaction({from: web3.eth.defaultAccount, to: "0x4e89c5FDf14877DA9757ea4c0FdC901023E5a868", value: web3.utils.toWei("3", "ether")})
+web3.eth.sendTransaction({from: web3.eth.defaultAccount, to: RELAYER_ADDRESS_GOES_HERE, value: web3.utils.toWei("3", "ether")})
 ```
 5. Relayer should register itself now that there's a stake. Otherwise restart the server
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -11,6 +11,31 @@
...
+```
+git clone https://github.com/tabookey/tabookey-gasless
+cd tabookey-gasless
+git checkout 2316c7422d50ac0242f8442f6dc98d0c85512c13
+npm install
+npm test
+```
+2. Run the server:
+```
+./build/server/bin/RelayHttpServer -RelayHubAddress 0x15d6e765aA9DaeBDdEB14141a35187FC850f87A3 -Workdir ./build/server --EthereumNodeUrl http://localhost:8555
```
jrainville: The localhost url means Embark needs to run right? If so, should probably specify it
Could you also specify what the HubAddress is. Is it ours? Or theirs?

README.md
```diff
@@ -11,6 +11,31 @@
...
+```
+Replace the RelayHub contract address for the correct address.
+3. Browse http://localhost:8090/getaddr. Copy the address
+4. Stake ether for that address. You can execute this in the embark console:
+```
+RelayHub.methods.stake("0x4e89c5FDf14877DA9757ea4c0FdC901023E5a868", 30).send({value: web3.utils.toWei("1", "ether"), gas:800000})
```
jrainville: We need to replace this address with the one we found? If so, can you put `...stake(THE_ADDRESS, 30)...`

contracts/teller-network/Escrow.sol
```diff
@@ -69,6 +79,63 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable {
...
+            if(transactions[escrowId].buyer != from) return 14;
+            if(metadataStore.getAsset(transactions[escrowId].offerId) != address(0)) return 15; // Must be eth trx
+
+            if(fSign == CANCEL_SIGNATURE){ // Allow activity after 15min have passed
+                if((lastActivity[from] + 15 minutes) > block.timestamp) return 17;
+            }
+        }
+
+        if(fSign == CREATE_SIGNATURE) {
+            // TODO: only allow eth transactions
```
jrainville: Do you still need that TODO?

contracts/teller-network/Escrow.sol
```diff
@@ -69,6 +79,63 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable {
...
 
+    mapping(address => uint) public lastActivity;
+
+    function accept_relayed_call(address relay, address from, bytes memory encoded_function, uint gas_price,
+      uint transaction_fee ) public view returns(uint32) {
+
+        bytes4 fSign = abiDecodeFunctionSignature(encoded_function);
+
+        if(fSign != CREATE_SIGNATURE && fSign != PAY_SIGNATURE && fSign != CANCEL_SIGNATURE && fSign != OPEN_CASE_SIGNATURE)
+            return 11;
```
jrainville: What happens when this returns not 0 (when it fails)? Does the relayer till pay gas or does it revert?

contracts/teller-network/Escrow.sol
```diff
@@ -69,6 +79,63 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable {
...
+            uint escrowId;
+            assembly {
+                escrowId := mload(add(escrowBytes, add(0x20, 0)))
+            }
+            if(escrowId >= transactions.length) return 13;
+            if(transactions[escrowId].buyer != from) return 14;
+            if(metadataStore.getAsset(transactions[escrowId].offerId) != address(0)) return 15; // Must be eth trx
+
+            if(fSign == CANCEL_SIGNATURE){ // Allow activity after 15min have passed
+                if((lastActivity[from] + 15 minutes) > block.timestamp) return 17;
```
jrainville: Why do we do that? So we block spammers?

naghdy:
➤ Nabil Naghdy commented:

Have now added this to Wrike Richard Ramos

README.md
richard-ramos: I'll add those items to the doc

contracts/teller-network/Escrow.sol
richard-ramos: anything different than 0 will not be relayed, and relayer does not pay gas nor anything. I guess they do a .call() first to see what will be the result of the call

contracts/teller-network/Escrow.sol
richard-ramos: yeah... it's a weak way to slow down spammers.

jrainville: Is there a way it could block real users too? Will we show a good message for it or will they just have a generic error?

richard-ramos: It will block real users too, so a good error message is required

PR #296, opened 2019-06-03
feat: show a message when the trade asset price is 10% lower

jrainville:
Percent can be easily changed
I'm not sure if I need to put the margin in my calculations? Like, what if the seller put amunally -10% 
margin, do we still put the warning?

Bogus data btw, ETH is not really 100 :P
image


jrainville:
Good idea. I guess only before funding. (Once it's funded, too late)
Also, should I show only to sellers?

richard-ramos:
I see it could be useful for buyers as well depending on price changes

PR #298, opened 2019-06-03
feat: hide token offers if you don't have ETH in your wallet

richard-ramos:


src/js/features/network/saga.js
```diff
@@ -17,7 +17,8 @@ export function *doInit() {
   try {
     yield call(onReady);
     const networkId = yield call(web3.eth.net.getId);
-    yield put({type: INIT_SUCCEEDED, networkId});
+    const gasPrice = yield web3.eth.getGasPrice();
+    yield put({type: INIT_SUCCEEDED, networkId, gasPrice});
```
jrainville: Here, you can just send the GET_PRICE action so that the other function will do it for you.
```
yield put({type: GET_GAS_PRICE});
```
You can even put it in a `all` to make it parallel.

src/js/pages/OffersList/index.jsx
```diff
@@ -86,7 +88,11 @@ class OffersList extends Component {
   };
 
   render() {
-    let filteredOffers = this.props.offers;
+    const toBN = web3.utils.toBN;
```
jrainville: You can put that `const` at the top of the file. That way we don't reinstantiate it every render

PR #300, opened 2019-06-04
feat: custom provider that only acts as relayer on low balance and calling specific functions

richard-ramos:


src/js/provider.js
```diff
@@ -0,0 +1,60 @@
...
+    send(payload, callback) {
+      const params = payload.params[0];
+      if(params && params.to && params.to.toLowerCase() === Escrow.options.address.toLowerCase()){
+        (async () => {
+          const balance = await web3.eth.getBalance(web3.eth.defaultAccount);
+          const gasPrice = await web3.eth.getGasPrice();
+          if(checkNotEnoughETH(gasPrice, balance) &&
+             payload.method === "eth_sendTransaction" &&
+             Object.values(VALID_OPERATIONS).includes(params.data.substring(2, 10))){
+              console.log("Relaying transaction");
```
jrainville: Probably not needed (the log)

src/js/provider.js
```diff
@@ -0,0 +1,60 @@
...
+
+    send(payload, callback) {
+      const params = payload.params[0];
+      if(params && params.to && params.to.toLowerCase() === Escrow.options.address.toLowerCase()){
+        (async () => {
+          const balance = await web3.eth.getBalance(web3.eth.defaultAccount);
+          const gasPrice = await web3.eth.getGasPrice();
+          if(checkNotEnoughETH(gasPrice, balance) &&
+             payload.method === "eth_sendTransaction" &&
+             Object.values(VALID_OPERATIONS).includes(params.data.substring(2, 10))){
```
jrainville: I'd check the metho and the operations first, and then check the balance and gasPrice, that way we only slow down by calling those if needed

PR #301, opened 2019-06-05
feat: disable cancel button when balance is not enough for transaction and last activity is < 15min

richard-ramos:


src/js/features/escrow/helpers.js
```diff
@@ -48,3 +48,13 @@ export function getTradeStatus(trade) {
     default: return tradeStates.waiting;
   }
 }
+
+const RELAY_DELAY = (15 * 60 * 1000) + 20; // Adding 20 seconds buffer since blocks are not mined exactly on this time
```
jrainville: That's 20 miliseconds, is this fine?

richard-ramos: lol! nop, it's not fine!  good catch!

PR #302, opened 2019-06-05
feat: disable escrow creation if it's going to be via relay and last activity < 15min

richard-ramos:


src/js/wizards/Buy/1_Trade/index.jsx
```diff
@@ -107,6 +110,14 @@ class Trade extends Component {
       return <Loading page/>;
     }
 
+    const notEnoughETH = checkNotEnoughETH(this.props.gasPrice, this.props.ethBalance);
+    const canRelay = escrow.helpers.canRelay(this.props.lastActivity);
+
+    let disabled = this.state.disabled;
+    if(notEnoughETH){
+      disabled = disabled && !canRelay;
```
jrainville: Shouldn't this be an `||`?

richard-ramos: no, because `disabled` contains the value of `this.state.disabled` which is set depending on the FIAT or Asset value. I want to take that logic in consideration. If I put `||`, if the transaction can be relayed, it will ignore any amount related restriction

jrainville: That's why I'm asking. With your change, if the  state is dasabled because of the FIAT or something, but it can relay, `disabled` will be false.

richard-ramos: ah, that's okay. I want the initial state of disabled to be taken in account first compared to the `canRelay` value.

PR #303, opened 2019-06-05
fix: reset create state 

richard-ramos:


richard-ramos:
Closing because I saw an issue with this PR

PR #311, opened 2019-06-13
feat: add fiat currencies handled by status app

richard-ramos:


src/js/wizards/Sell/5_Currency/index.jsx
```diff
@@ -6,18 +6,7 @@ import {connect} from 'react-redux';
...
-//       and it does not have a full list of FIAT currencies
-const CURRENCY_DATA = [
-  {id: 'USD', label: 'United States Dollar - USD'},
-  {id: 'EUR', label: 'Euro - EUR'},
-  {id: 'GBP', label: 'Pound sterling - GBP'},
-  {id: 'JPY', label: 'Japanese Yen - JPY'},
-  {id: 'CNY', label: 'Chinese Yuan - CNY'},
-  {id: 'KRW', label: 'South Korean Won - KRW'}
-];
+import {CURRENCY_DATA} from "../../../utils/currencies";
```
andremedeiros: Shouldn't this live in constants instead of utils?

richard-ramos: We do not have a constants folder.  might be a good moment to create one!

PR #314, opened 2019-06-14
feat: decouple metadata from escrow, and ask for signatures in escrow creation

richard-ramos:
Stops metadata from depending on the Escrow contract.
This includes a flow change:
1. Signature is being asked in the buyer flow after the username and contact code screen.
2. Removes the redirect on the username/contact code screen, due to this signature requiring a nonce.

### DIFF
--- a/contracts/teller-network/MetadataStore.sol
+++ b/contracts/teller-network/MetadataStore.sol
@@ -1,14 +1,13 @@
 pragma solidity ^0.5.8;
 
 import "./License.sol";
-import "../common/Ownable.sol";
 import "../common/MessageSigned.sol";
 
 /**
 * @title MetadataStore
 * @dev Metadata store
 */
-contract MetadataStore is Ownable, MessageSigned {
+contract MetadataStore is MessageSigned {
 
     enum PaymentMethods {Cash,BankTransfer,InternationalWire}
 
@@ -80,11 +79,6 @@ contract MetadataStore is Ownable, MessageSigned {
         license = _license;
         arbitrationLicense = _arbitrationLicense;
     }
-
-    function setEscrowAddress(address _escrow) public onlyOwner {
-        escrow = _escrow;
-    }
-
 
     mapping(address => uint) public user_nonce;
 
--- a/contracts/teller-network/MetadataStore.sol
+++ b/contracts/teller-network/MetadataStore.sol
@@ -171,7 +171,6 @@ contract MetadataStore is Ownable, MessageSigned {
     */
     function addOffer(
         address _asset,
-        bytes memory _signature,
         bytes memory _statusContactCode,
         string memory _location,
         string memory _currency,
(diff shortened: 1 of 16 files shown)


src/js/wizards/Buy/0_Contact/index.jsx
```diff
@@ -22,21 +22,19 @@ class Contact extends Component {
...
-      // FIXME: infinite loop between this page and the next
-      setTimeout(() => {
-        this.props.wizard.next();
-      }, 500);
     } else {
       this.validate(this.props.username, this.props.statusContactCode);
-      this.setState({ready: true});
     }
+
+    this.setState({ready: true});
```
jrainville: Ready doesn't really achieve much now. 
If we still want a loading, we can just put an `if` in the render and show the loading only `if (!this.props.profile)`

test/escrow_spec.js
```diff
@@ -109,32 +108,48 @@ contract("Escrow", function() {
...
+        await Escrow.methods.create(signature, ethOfferId, 123, FIAT, 140, "0x00", "L", "U", nonce).send({from: accounts[8]});       
         assert.fail('should have reverted before');
       } catch (error) {
         assert.strictEqual(error.message, "VM Exception while processing transaction: revert Must participate in the trade");
       }
     });
 
     it("Buyer can create escrow", async () => {
-      receipt = await Escrow.methods.create(accounts[1], ethOfferId, 123, FIAT, 140, [0], "L", "U").send({from: accounts[1]});
+      // receipt = await Escrow.methods.create(accounts[1], ethOfferId, 123, FIAT, 140, [0], "L", "U").send({from: accounts[1]});
```
jrainville: 🗑

contracts/teller-network/MetadataStore.sol
```diff
@@ -80,28 +79,69 @@ contract MetadataStore is Ownable {
...
-        address _user,
         bytes memory _statusContactCode,
         string memory _location,
         string memory _username
     ) public {
-        if(msg.sender != escrow){
-            require(msg.sender == _user, "Sender does not match address");
-        }
+        _addOrUpdateUser(msg.sender, _statusContactCode, _location, _username);
+    }
```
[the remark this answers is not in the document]
richard-ramos: the differences are:
- `function addOrUpdateUser( bytes memory _signature, bytes memory _statusContactCode,         string memory _location, string memory _username, uint _nonce)` will obtain the user address from the signature and call `_addOrUpdateUser`
- `function addOrUpdateUser( bytes memory _statusContactCode,         string memory _location, string memory _username, uint _nonce)`  since it doesn't receive an user, it will use `msg.sender` when calling `_addOrUpdateUser`.  It's mostly for convenience, since calling this function will not require obtaining a signature, if the user to be added/updated is the msg.sender.

PR #316, opened 2019-06-17
feat: offers can be marked as deleted in the contract, and make the UI filter these

richard-ramos:


richard-ramos:
yeah, basically it's for cases where the escrow is open, and also to not lose the history for escrows already done with this offer. 
I'll add the logic to check if an offer is deleted or not when creating new escrows

PR #323, opened 2019-06-20
Implement 1 percent fees

jrainville:
Changes the fee system to be a percentage of the current amount instead of a flat SNT fee.
This required a lot of changes, because we no longer approve SNT all the time, but only for SNT trades.
Also, the fee in the contract is in milli-percents because decimals don't exist, so this enables us to use smaller values than 1 percent.
Please review thoroughly since this is the first time I change contracts that much.


test/license_spec.js
```diff
@@ -70,11 +66,12 @@ contract("License", function () {
...
     await SNT.methods.approve(License.options.address, 10).send({from: accounts[0]});
     await License.methods.buy().send({from: accounts[0]});
-    
+
     isLicenseOwner = await License.methods.isLicenseOwner(accounts[0]).call();
     assert.strictEqual(isLicenseOwner, true);
-    const stakingBalance = await SNT.methods.balanceOf(StakingPool.options.address).call();
-    assert.strictEqual(stakingBalance, "10", "Contract balance is incorrect");
+    // const stakingBalance = await SNT.methods.balanceOf(StakingPool.options.address).call();
+    // FIXME This test doesn't pass even on master
```
richard-ramos: There's a PR pending review that fixes this test

contracts/teller-network/Fees.sol
```diff
@@ -36,45 +34,66 @@ contract Fees is Ownable {
...
 
     /**
      * @notice Withdraw fees by sending them to the fee destination address
      */
-    function withdrawFees() public {
-        uint fees = feeBalance;
-        feeBalance = 0;
-        require(feeToken.transfer(feeDestination, fees), "Error transfering fees");
-        emit FeesWithdrawn(fees);
+    function withdrawFees(address _tokenAddress) public {
```
richard-ramos: Are we still going to hold the fees in the escrow contract until this function is called? or should we send the fees immediatly to the stake contract?

contracts/teller-network/Fees.sol
```diff
@@ -36,45 +34,66 @@ contract Fees is Ownable {
...
      * @param _from Address from where the fees are being extracted
      * @param _id Escrow id or element identifier to mark as paid
+     * @param _value Value sold in the escrow
+     * @param _feeAmount Fee amount calculated by the front-end
+     * @param _tokenAddress Address of the token sold in the escrow
      * @dev This will only transfer funds if the fee  has not been paid
      */
-    function payFee(address _from, uint _id) internal {
-        if(feePaid[_id]) return;
+    function payFee(address _from, uint _id, uint _value, uint _feeAmount, address _tokenAddress) internal {
```
richard-ramos: I'm confused on why the _feeAmount must be an input value, instead of the contract just receiving an amount and doing the fee split on its own.

contracts/teller-network/Fees.sol
```diff
@@ -36,45 +34,66 @@ contract Fees is Ownable {
...
         feePaid[_id] = true;
-        feeBalance += feeAmount;
+        uint amount = getFeeFromAmount(_value);
+        require(amount == _feeAmount, "Fee amount needs to match the percentage");
+        feeTokenBalances[_tokenAddress] += amount;
 
-        require(feeToken.transferFrom(_from, address(this), feeAmount), "Unsuccessful token transfer");
+        if (_tokenAddress != address(0)) {
+            ERC20Token tokenToPay = ERC20Token(_tokenAddress);
+            require(tokenToPay.allowance(_from, address(this)) >= amount, "Allowance not set for this contract for specified fee");
```
richard-ramos: In a pending PR I'm removing the need of checking allowances, since we already do a require that verifies if the transfer is successful.  This line can be removed.

contracts/teller-network/Fees.sol
jrainville: I didn't code the later part of the fees, ie I didn't do the part where we send to the arbiter or the staking pool yet because I wanted to get the first part out before it got more complex.

contracts/teller-network/Fees.sol
jrainville: Yea....... I think you're right and I don't need it at all haha
At first, I thought I needed to pass it so that the Fees contract could make sure the front-end sent the right amount, but it's all in the allowance or the value passed, so it will already be reverted if the user didn't allow or pass enough value.
Good catch, will change that

jrainville:
Updated @richard-ramos

src/js/features/escrow/saga.js
```diff
@@ -28,6 +28,9 @@ import {ADD_OFFER_SUCCEEDED} from "../metadata/constants";
 
 const { toBN } = web3.utils;
 
+window.MetadataStore = MetadataStore;
```
richard-ramos: This is for the trick of creating escrows without requiring using metamask, right?

jrainville: Oops, yes, my bad.

PR #325, opened 2019-06-25
refactor: escrow instances

richard-ramos:
New contracts:
1. EscrowManagement: serves as factory for escrows
2. EscrowRelay: contract that can perform relay operations on escrow
3. Proxy + ProxyData: proxy logic for escrows
4. Arbitrations: logic for handling disputes for escrow instances

Almost all the contract logic was affected, so this PR is against a branch in order to have a separate PR later for the UI changes.


contracts/teller-network/Escrow.sol
```diff
@@ -161,55 +157,29 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable, RelayRecipient {
...
      */
-    function fund(uint _escrowId, uint _tokenAmount, uint _expirationTime) external payable whenNotPaused {
-        _fund(get_sender(), _escrowId, _tokenAmount, _expirationTime);
+    function fund(uint _tokenAmount, uint _expirationTime) external payable {
+        _fund(msg.sender, _tokenAmount, _expirationTime);
     }
 
-    function _fund(address _from, uint _escrowId, uint _tokenAmount, uint _expirationTime) internal whenNotPaused {
+    function _fund(address _from, uint _tokenAmount, uint _expirationTime) internal {
+        // TODO: expiration time
```
jrainville: What's left to do with the expiration?

contracts/teller-network/Escrow.sol
```diff
@@ -218,327 +188,256 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable, RelayRecipient {
...
-        address seller;
-        address arbitrator;
+    ) internal {
         bool deleted;
 
-        (,,,,seller, arbitrator, deleted) = metadataStore.offer(_offerId);
+        (token, , , , seller, arbitrator, deleted) = metadataStore.offer(_offerId);
 
         require(!deleted, "Offer is not valid");
-        require(get_sender() == _buyer || get_sender() == seller, "Must participate in the trade");
```
jrainville: Why remove this check?

contracts/teller-network/Escrow.sol
```diff
@@ -218,327 +188,256 @@ contract Escrow is Pausable, MessageSigned, Fees, Arbitrable, RelayRecipient {
...
-     * @dev Consider using Aragon Court for this.
      */
-    function openCase(uint _escrowId, string calldata motive) external {
-        EscrowTransaction storage trx = transactions[_escrowId];
-
-        require(!isDisputed(_escrowId), "Case already exist");
-        require(trx.buyer == get_sender() || metadataStore.getOfferOwner(trx.offerId) == get_sender(), "Only a buyer or seller can open a case");
-        require(trx.status == EscrowStatus.PAID, "Cases can only be open for paid transactions");
+    function openCase(string calldata _motive) external {
+        require(buyer == msg.sender || seller == msg.sender, "Only a buyer or seller can open a case");
```
jrainville: Should we create a modifier for this, we reuse it a lot, maybe might save some gas to deploy the contract

contracts/teller-network/EscrowRelay.sol
```diff
@@ -0,0 +1,202 @@
...
+    bytes memory encoded_function,
+    uint gas_price,
+    uint transaction_fee
+  ) public view returns(uint32) {
+    bytes4 fSign;
+    assembly {
+      fSign := mload(add(encoded_function, add(0x20, 0)))
+    }
+
+    if(from.balance > 600000 * gas_price) return ERROR_ENOUGH_BALANCE;
```
jrainville: Did you increase this? Since creating an escrow is now more costly?

test/arbitration_spec.js
```diff
@@ -0,0 +1,305 @@
...
+      receipt = await EscrowManagement.methods.create(ethOfferId, tradeAmount, FIAT, 140, "0x00", "L", "U", nonce, signature).send({from: accounts[1]});
+      Escrow.options.address = receipt.events.InstanceCreated.returnValues.instance;
+      // Fund
+      receipt = await Escrow.methods.fund(tradeAmount, expirationTime).send({from: accounts[0], value: tradeAmount + feeAmount});
+    });
+
+    it("should allow a buyer to open a case", async() => {
+      await Escrow.methods.pay().send({from: accounts[1]});
+
+      Escrow.options.jsonInterface.push(Arbitrations.options.jsonInterface.find(x => x.name === "ArbitrationRequired"));
```
jrainville: Is this because the Arbitration methods are not present in the Escrow json interface?

contracts/teller-network/Escrow.sol
richard-ramos: Currently it's handled as an input parameter. I think we should change this so the expiration time is calculated by the contract instead.

contracts/teller-network/Escrow.sol
richard-ramos: Because now since the buyers are required to sign a message to create an escrow, they will always be a participant on the trade; and this signature is unique due to it containing an user nonce that increments on each trx, so there is no way of creating a escrow without the buyer participating in the trade

contracts/teller-network/EscrowRelay.sol
richard-ramos: Good catch. I did it on #327

test/arbitration_spec.js
richard-ramos: I did it because `Escrow` will call `Arbitrations.openCase`, and the `ArbitrationRequired` event will be emitted there, but since `Escrow` does not know the details of that event, I added it to its interface in order to be able to decode that log.

Maybe we should add a configuration attribute to the contracts config in Embark to simplify this procedure? something like:
```
contracts {
     "Arbitrations":  { ... },
     "Escrow": {
        ...,
        "inheritEventsFrom": ["Arbitrations"]
     }
}
```
what do you think?

contracts/teller-network/Escrow.sol
richard-ramos: Good idea!

test/arbitration_spec.js
jrainville: Yeah, to be fair, it would be even cooler if Embark detected inheritance and put all the events of the parents in the json interface, but `inheritEventsFrom` could be a good a good alternative

richard-ramos:
Closing this PR because instead of having multiple escrow instances, we'll have a single upgradable contract. PR #332 does this.

PR #326, opened 2019-06-25
Send fees to the arbiter and /or the Staking pool

jrainville:
Send 10% of the fee to the arbiter when no dispute, and 100% when there is a dispute.

### DIFF
--- a/contracts/teller-network/Fees.sol
+++ b/contracts/teller-network/Fees.sol
@@ -16,7 +16,6 @@ contract Fees is Ownable {
     event FeeDestinationChanged(address payable);
     event FeeMilliPercentChanged(uint amount);
     event FeesWithdrawn(uint amount, address token);
-    event MyEvent(uint fees, address destination, uint balance);
 
     /**
      * @param _feeDestination Address to send the fees once withdraw is called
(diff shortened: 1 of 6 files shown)


contracts/teller-network/Fees.sol
```diff
@@ -18,6 +18,7 @@ contract Fees is Ownable {
     event FeesWithdrawn(uint amount, address token);
     event MyEvent(uint fees, address destination, uint balance);
 
+    event MyEvent(address from, uint allowance, uint tokenAmount, uint destinationValue, uint feeAmount);
```
richard-ramos: leftovers from debugging?

contracts/teller-network/Fees.sol
```diff
@@ -51,26 +52,47 @@ contract Fees is Ownable {
...
+     * @param _arbitrator Arbitrator address to transfer fee to
+     * @param _value Value sold in the escrow
+     * @param _isDispute Boolean telling if it was from a dispute. With a dispute, the arbitrator gets more
+    */
+    function releaseFee(address payable _arbitrator, uint _value, address _tokenAddress, bool _isDispute) internal {
+        uint _milliPercentToArbitrator;
+        if (_isDispute) {
+            _milliPercentToArbitrator = 100000; // 100%
+        } else {
+            _milliPercentToArbitrator = 10000; // 10%
```
richard-ramos: Was the 10% for arbitrators even if they do nothing the final decision?

contracts/teller-network/Fees.sol
```diff
@@ -51,26 +52,47 @@ contract Fees is Ownable {
...
+
+        uint feeAmount = getValueOffMillipercent(_value, feeMilliPercent);
+        uint arbitratorValue = getValueOffMillipercent(feeAmount, _milliPercentToArbitrator);
+        uint destinationValue = feeAmount - arbitratorValue;
+
+        if (_tokenAddress != address(0)) {
+            ERC20Token tokenToPay = ERC20Token(_tokenAddress);
+            // Arbitrator transfer
+
+            uint allow = tokenToPay.allowance(address(this), address(this));
```
richard-ramos: `allow` and `myEvent` are probably part of a debugging session and should be removed

contracts/teller-network/Fees.sol
jrainville: Yes, arbiters get 10% of the fee when they do nothing, that was what was decided. It can be changed obviously.
Also, it ends up being 10% of 1%, so it's 0.1% of the amount

jrainville:
Updated and rebased

PR #328, opened 2019-06-26
refactor: Provider will determine if it needs to use the EscrowRelay contract

richard-ramos:


richard-ramos:
oops. wrong branch

PR #329, opened 2019-06-26
refactor: make provider determine if it needs to use EscrowRelay contract

richard-ramos:


src/js/wizards/Buy/0_Contact/index.jsx
```diff
@@ -75,6 +79,7 @@ class Contact extends Component {
   }
 
   render() {
+    console.log(this.state);
```
jrainville: stray console

richard-ramos:
Closing this PR since most of these changes are not required when we only use a single upgradable escrow

PR #324, opened 2019-06-24
add: Arbitrator licensing and seller management

arsenyjin:

PR #337, opened 2019-07-01
add: arbitration selections checks, fixes, clean up

arsenyjin:
Includes code and recommendation from https://github.com/status-im/status-teller-network/pull/324


contracts/teller-network/Escrow.sol
```diff
@@ -1,4 +1,4 @@
-/* solium-disable security/no-block-members */
+ /* solium-disable security/no-block-members */
```
jrainville: Small space added by accident

PR #342, opened 2019-07-03
feat: documentation and code format

richard-ramos:


contracts/common/Controlled.sol
```diff
@@ -1,17 +1,17 @@
-pragma solidity ^0.5.7;
+pragma solidity >=0.5.0 <0.6.0;
```
andremedeiros: 🙏

contracts/teller-network/MetadataStore.sol
```diff
@@ -99,7 +99,7 @@ contract MetadataStore is MessageSigned {
         uint _nonce,
         bytes memory _signature
     ) internal view returns(address) {
-        return recoverAddress(getSignHash(_dataHash(_username, _statusContactCode, _nonce)), _signature);
+        return _recoverAddress(_getSignHash(_dataHash(_username, _statusContactCode, _nonce)), _signature);
```
andremedeiros: Can this be a couple more lines, but with temporary variables so that it's easier to read?

PR #346, opened 2019-07-04
fix: approval of tokens

jrainville:
SNT escrows went straight to funding instead of  approval.
Also, I'm pretty sure the approval was bugged, because we were approving only "ether" amounts and not wei amounts... so I'm not sure why it worked before. Just confirm that I'm doing the right thing and we'll be able to test later if the fees and everything is fine

### DIFF
--- a/src/js/features/approval/saga.js
+++ b/src/js/features/approval/saga.js
@@ -3,6 +3,7 @@
 import {fork, put, takeEvery} from 'redux-saga/effects';
 import {doTransaction} from '../../utils/saga';
 import {zeroAddress} from '../../utils/address';
+import {toTokenDecimals} from '../../utils/numbers';
 import SNT from '../../../embarkArtifacts/contracts/SNT';
 import Escrow from '../../../embarkArtifacts/contracts/Escrow';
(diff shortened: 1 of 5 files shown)


src/js/features/approval/saga.js
```diff
@@ -10,14 +10,15 @@ import { APPROVE_TOKEN, APPROVE_PRE_SUCCEEDED, APPROVE_SUCCEEDED, APPROVE_FAILED
 import OwnedUpgradeabilityProxy from '../../../embarkArtifacts/contracts/OwnedUpgradeabilityProxy';
 Escrow.options.address = OwnedUpgradeabilityProxy.options.address;
 
-const {toBN} = web3.utils;
+const {toBN, toWei} = web3.utils;
 
 export function *approveToken({amount}) {
   const feeMilliPercent = yield Escrow.methods.feeMilliPercent().call();
+  const amountWei = toBN(toWei(amount));
```
richard-ramos: shouldnt we use toTokenDecimals in case the token doesn't use 18 decimals?

jrainville: Ah yes, forgot about that

jrainville: Updated

PR #351, opened 2019-07-04
feat: sellers' arbitration selection screen

richard-ramos:


src/js/features/arbitration/reducer.js
```diff
@@ -153,11 +168,36 @@ function reducer(state = DEFAULT_STATE, action) {
...
+        return {
+          ...state,
+          arbitrators,
+          loading: false,
+          error: ''
+        };
+      }
+    case CANCEL_ARBITRATOR_REQUEST_SUCCEEDED: 
+      {
+        console.log(action);
```
jrainville: 🔪

src/js/pages/Arbitrators/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+class Arbitrators extends Component {
+  constructor(props) {
+    super(props);
+    props.getArbitrators(props.address, true);
+    this.loadedUsers = [];
+
+  }
+
+  componentDidUpdate(prevProps) {
+    if ((!prevProps.arbitrators && this.props.arbitrators) || prevProps.arbitrators.length !== this.props.arbitrators.length || Object.keys(this.props.users).length !== this.props.arbitrators.length) {
```
jrainville: Not sure the last part is ok. ` Object.keys(this.props.users).length !== this.props.arbitrators.length`
I think it should be ` Object.keys(this.props.users).length < this.props.arbitrators.length`
because there will most likely be more users than arbitrators

richard-ramos: good catch!

PR #364, opened 2019-07-09
fix: preload buyer contact data and enable next button

richard-ramos:


src/js/wizards/Buy/0_Contact/index.jsx
```diff
@@ -40,12 +40,8 @@ class Contact extends Component {
       this.changeStatusContactCode(this.props.apiContactCode);
     }
 
-    if (prevProps.statusContactCode !== this.props.statusContactCode) {
-      this.changeStatusContactCode(this.props.statusContactCode);
-    }
-
-    if (prevProps.username !== this.props.username) {
-      this.changeUsername(this.props.username);
+    if (prevProps.statusContactCode !== this.props.statusContactCode && prevProps.username !== this.props.username) {
```
jrainville: Any chance that statusContactCode  and username are changed at different times?

richard-ramos: Not sure. I did test assuming it's the first time the user creates a escrow; after an user already has an escrow, and also by creating an escrow after the profile is updated and seems to behave correctly

PR #375, opened 2019-07-16
Fix solc warnings and update embark-solium

jrainville:
I think embark-solium never actually worked. It was also using an older version

### DIFF
--- a/contracts/teller-network/EscrowRelay.sol
+++ b/contracts/teller-network/EscrowRelay.sol
@@ -151,6 +151,8 @@ contract EscrowRelay is RelayRecipient, Ownable {
    * @param encoded_function Function that will be called on the Escrow contract
    * @param gas_price Gas price
    * @param transaction_fee Fee for the relay (unused by us)
+   * @dev relay and transaction_fee give warning because they are unused, but they are useless in our relay workflow
+   * @dev We cannot remove those parameters because they are called by an external contract
    */
   function accept_relayed_call(
     address relay,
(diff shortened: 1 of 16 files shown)


contracts/teller-network/EscrowRelay.sol
```diff
@@ -89,7 +89,7 @@ contract EscrowRelay is RelayRecipient, Ownable {
    * @param _nonce buyer's nonce
    * @param _signature buyer's signature
    */
-  function create (
+  function createEscrow(
```
richard-ramos: the value of `CREATE_SIGNATURE` must change from create(....)  to createEscrow(....)

jrainville: Ah, yes, good catch

contracts/teller-network/IEscrow.sol
```diff
@@ -17,7 +17,7 @@ contract IEscrow {
       address payable arbitrator;
   }
 
-  function create(
+  function createEscrow(
```
richard-ramos: Changing from `create` to `createEscrow` is required in the saga too.

jrainville: 🤦‍♂ of course, my bad.

contracts/teller-network/EscrowRelay.sol
```diff
@@ -164,6 +164,10 @@ contract EscrowRelay is RelayRecipient, Ownable {
       fSign := mload(add(encoded_function, add(0x20, 0)))
     }
 
+    if (relay == address(0) || transaction_fee == 0) {
```
richard-ramos: Can we ignore this warning?... just to avoid spending gas... probably it's not even much, but even if this function is `view`, it is used in `RelayHub`'s `relay` function.

jrainville: I tried to search and it didn't seem possible to do so. The only suggestion the solc maintainer said was to comment it, but we probably can't because it's not us calling the function?
Maybe there is a better way to seem like we use the variable without doing anything?

richard-ramos: The other option is to accept that this warning will exists and add a comment explaining that those variables aren't necessary in our relayer conditions.

jrainville: Yeah, makes sense

PR #387, opened 2019-07-22
fix: escrow warnings for buyers and sellers

richard-ramos:


richard-ramos:
I changed the diff percentage calculation to take in account if it's the seller or the buyer the one looking at the escrow details:

Seller:
Screenshot1

Buyer:
Screenshot2

however, I'm not sure if the warning should be displayed for buyers when the current price of the selected token is above the price of the escrow;  or sellers when the price of the escrow is above the price in the offer...

PR #388, opened 2019-07-22
Blacklist sellers for arbiters accepting everyone

jrainville:
Let's arbiters who accept anyone still blacklist bad boys and girls

image


jrainville:
> What do you think about making blacklists available to arbitrators that don't accept all sellers in a future iteration?

We could, but I don't feel like it adds much apart from a future proofing.

Explanation: If you don't use `Accept all`, you basically use a whitelist, so even if you blacklist someone, you're just saying, I REALLY don't want to accept that user.
The only use to blacklist while **not** using `Accept all` is IF you change to `Accept all` later, then you ave your blacklisted sellers already.

Do you think it's worth it? I mean, it's not a big change and I purposefully didn't restrict it the contract.

PR #390, opened 2019-07-23
fix: reduce escrow creation gas cost by 50K

richard-ramos:


contracts/teller-network/Arbitrable.sol
```diff
@@ -47,14 +47,18 @@ contract Arbitrable {
...
     /**
      * @notice Determine if a dispute exists/existed for an escrow
      * @param _escrowId Escrow to verify
      * @return bool result
      */
     function isDisputed(uint _escrowId) public view returns (bool) {
+        return _isDisputed(_escrowId);
+    }
+
+    function _isDisputed(uint _escrowId) internal view returns (bool) {
```
jrainville: How does this make it cost less?

richard-ramos: Changing this function to internal is translated into a jump inside the EVM: when you call the public function like I did before on `Escrow.openCase`, all the functions parameters are copied in memory and passed to the function. When it was changed to internal, references to the parameters are passed and not copied in memory again, which saves some gas

jrainville: Cool, thanks for the explanation

PR #392, opened 2019-07-24
fit: split pubkey into x/y coordinates

richard-ramos:


embarkConfig/data.js
```diff
@@ -144,7 +144,7 @@ module.exports = async (licensePrice, arbitrationLicensePrice, feeMilliPercent,
         tokens[1],
         // TODO un hardcode token and add `approve` in the escrow creation below
         // tokens[Math.floor(Math.random() * tokens.length)],
-        address,
+        address + "112233445566778899001122",
```
jrainville: This is a random number I guess

richard-ramos: yup... just to have a valid bytes32 pubkey

PR #404, opened 2019-07-30
feat: calculate limits for buyers

richard-ramos:


src/js/wizards/Buy/1_Trade/components/OfferTrade.jsx
```diff
@@ -52,6 +47,16 @@ const OfferTrade = ({
...
           </Row>
         </FormGroup>
+        { limitless && <p className="mt-3">
+        Limits: {minToken} {asset} to <span id="max-token">{maxToken} {asset}</span>
+        <UncontrolledTooltip placement="right" target="max-token">
+          This is the current balance of the seller. This is why it is the maximum
+        </UncontrolledTooltip>
+        </p> }
+        { !limitless && <p className="mt-3">
+        Limits: {(parseFloat(limitL) / 100).toFixed(2)} {currency.id} to <span id="max-token">{(parseFloat(limitH) / 100).toFixed(2)} {currency.id}</span>
```
jrainville: Here, we should check if the seller's balance is not sufficient. Maybe just add a warning saying that the seller currently doesn't have the available funds.

richard-ramos: Good idea!

PR #405, opened 2019-07-31
feat: show no license page for users that are not sellers and attempt to access a seller module

richard-ramos:


src/js/components/NoLicense/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+const NoLicense = ({isArbitrator}) => (
+<div className="no-license">
+    {isArbitrator && <h3>You don&apos;t have an arbitrator license </h3>}
+    {!isArbitrator && <h3>You don&apos;t have a seller license </h3>}
+    {isArbitrator && <p className="text-muted">Once you are approved as an arbitrator you can make tokens by judging disputes.</p> }
+    {!isArbitrator && <p className="text-muted">Once you are a seller you will be able to create offers</p> }
+    <p className="mt-5 pt-5 text-center">
+        {isArbitrator && <Button color="primary" tag={Link} to="/arbitrator/license" className="mt-5">Become an arbitrator</Button> }
+        {!isArbitrator && <Button color="primary" tag={Link} to="/license" className="mt-5">Become a seller</Button> }
```
jrainville: Actually, all conditions here seem wrong. Am I missing something? Shouldn't `isSeller` be included too?

jrainville: AHHH, I just checked the other  component and now I understand that it's supposed to be about the type of license, but currently, `isArbitrator ` makes it seems like the user is an arbitrator.
I would rename the property to `isArbitratorPage` or something like that

PR #409, opened 2019-07-31
feat: calculate limits for buyers

richard-ramos:


src/js/wizards/Buy/1_Trade/components/OfferTrade.jsx
```diff
@@ -10,19 +10,18 @@ import moment from "moment";
...
-  assetQuantity, currencyQuantity, onCurrencyChange, onAssetChange, disabled, t, notEnoughETH, canRelay
-}) => (
-  <Row>
+  statusContactCode, name, minToken, maxToken, price, currency, asset, lastActivity, limitless,
+  assetQuantity, currencyQuantity, onCurrencyChange, onAssetChange, disabled, t, notEnoughETH, canRelay,
+  limitH, limitL, sellerBalance
+}) => {
+  const minFiat = (parseFloat(limitL) / 100).toFixed(2);
+  const maxFiat = (parseFloat(limitH) / 100).toFixed(2);
+  const amountGreaterThanBalance = currencyQuantity > parseFloat(sellerBalance) * price;
```
jrainville: Not sure why you're comparing the fiat amounts, though it probably gives the same result.
Comparing sellerBalance and assetQuantity would have been simpler I think

src/js/wizards/Buy/1_Trade/index.jsx
```diff
@@ -56,11 +56,22 @@ class Trade extends Component {
   }
 
   validate(currencyQuantity, assetQuantity) {
-    if (currencyQuantity < 0 || assetQuantity < 0 || assetQuantity > (this.props.sellerBalance || MAX)) {
-      this.props.footer.disableNext();
-      this.setState({disabled: true});
-      return;
+    const limitless = this.props.offer.limitL === '0' && this.props.offer.limitH === '0';
+    if(limitless){
+      if (currencyQuantity < 0 || assetQuantity < 0 || assetQuantity > (this.props.sellerBalance || MAX)) {
```
jrainville: Don't know if we should disable on quantity higher than seller balance? Could still open and then let seller restock. What do you think?

PR #412, opened 2019-08-01
feat: styling for trades

richard-ramos:


src/js/pages/MyDisputes/index.scss
```diff
@@ -19,3 +19,7 @@
   border: 1px solid #EEF2F5;
   float: right;
 }
+
+.no-border {
```
jrainville: Bootstrap already has that. It's `border-0`

src/js/pages/MyTrades/components/Trades.scss
```diff
@@ -1,5 +1,15 @@
+@import "../../../../css/variable-overrides";
+
 .state-filters {
   .form-control {
     height: 40px;
   }
 }
+
+.no-border {
```
jrainville: Same here, no needed since `border-0` exists

src/js/pages/MyTrades/components/Trades.scss
```diff
@@ -1,5 +1,15 @@
...
+  border: 0;
+}
+
+.profile-trades-list a {
+  color: $black;
```
jrainville: You can use `.text-black` if you want. Though, I get that it'S simpler to target all links from the css.

If the `<a>`s have a class (eg `link`), it's better to do `.profile-trades-list .link {` because css is interpreted from right to left and it's a bit slower to use tags as the right-most element in a declaration.

PR #422, opened 2019-08-06
feat: updating GSN to latest version

richard-ramos:


contracts/teller-network/EscrowRelay.sol
```diff
@@ -140,79 +140,95 @@ contract EscrowRelay is RelayRecipient, Ownable {
...
+    } else if(functionSignature == CREATE_SIGNATURE) {
+      token = metadataStore.getAsset(dataValue);
 
       if(token != snt && token != address(0)) return ERROR_INVALID_ASSET;
 
       // Allow activity after 15 min have passed
       if((lastActivity[from] + 15 minutes) > block.timestamp) return ERROR_TRX_TOO_SOON;
     }
-
     return OK;
```
jrainville: What happens if the functionSignature is for `createOffer`, but the suer has no ETH, would it say OK?

richard-ramos: `createOffer` would not be a valid function to be executed with the relayer. Only the public functions of EscrowRelay can be relayed due to it being a RelayRecipient... but I'll change it to return OK only inside the `if`s   in case another function of this contract were to be executed via relay and it was not intended to.

jrainville: Yeah, you're right, but I agree that we should only return OK for the wanted functions

PR #423, opened 2019-08-07
fix: fix footer to bottom and remove duplicated payment methods

richard-ramos:


src/index.scss
```diff
@@ -17,6 +17,15 @@ body {
...
     }
+
+    .wizard-container .footer {
+      position: fixed;
+      background: #fff;
+      bottom: 0;
+      z-index: 9999;
+      padding-top: 10px;
+      padding-bottom: 10px;
+    }
```
jrainville: I think we have a style sheet for the footer already.

Also, is it ugly in other pages to fix it? I think I remember coding the footer as I did because it was better to have it follow the structure but stick sometimes

richard-ramos: I changed it for all the pages because I noticed in the design that the footer was always fixed to bottom regardless of the content height.

jrainville: Alright, it's fine. Just maybe change the style in the footer style sheet instead

PR #427, opened 2019-08-08
Fix/disabled button css

richard-ramos:


src/index.scss
```diff
@@ -31,7 +31,12 @@ body {
 
   .clickable {
     cursor: pointer;
+    
+    .disabled {
```
jrainville: Are you sure you don't mean `&.disabled`?

richard-ramos: you're right!

richard-ramos: ah no. the disabled is applied to childs of .clickable :)

jrainville: Ah ok, I trust you :)

PR #430, opened 2019-08-13
fix: filter v1 UI bugs

richard-ramos:
Still missing the "Best price" filter. I'm still not sure how to implement this


richard-ramos:
I guess I'll do that in a separate PR

PR #431, opened 2019-08-13
UI fixes for escrow page and others

jrainville:


src/js/components/Reputation/index.jsx
```diff
@@ -26,11 +26,11 @@ class Reputation extends Component {
       let classes = 'average-rating py-1 px-3 rounded ';
       if (!Number.isInteger(reputation.averageCount)) {
         classes += 'bg-dark text-dark';
-      } else  if (reputation.averageCount >= 4) {
+      } else  if (reputation.averageCount >= 8) {
```
richard-ramos: We have a condition in the contracts to not allow ratings above 5, so probably most of these conditions will not trigger.

richard-ramos: Nvm. I see you convert it below to base10!  clever!

jrainville: Yeah, it,s not ideal, but that way, we don't have to modify the contract haha

PR #433, opened 2019-08-13
fix: ui styles for offer creation

richard-ramos:


src/js/wizards/Sell/3_Location/components/SellerPosition.jsx
```diff
@@ -10,13 +10,13 @@ class SellerPosition extends Component {
   render() {
     return (
       <React.Fragment>
-        <h2 className="mb-4">What location do you want to display</h2>
+        <h2>What location do you want to display</h2>
         <Form>
           <FormGroup>
-            <Input type="text" name="location" id="location" placeholder="City, Country"
+            <Label className="text-small mt-3 mb-0">City</Label>
+            <Input type="text" name="location" id="location" placeholder="Enter City"
```
jrainville: Since we don't have a city selector for now, maybe we should keep it at `Location` or something instead of City

PR #438, opened 2019-08-16
feat: implement new branding with new logo and colour

jrainville:
To change the color of the logs, in the end, I just had to replace in path the fill to the new one

image


.env
```diff
@@ -0,0 +1 @@
+HOST=0.0.0.0
```
richard-ramos: :question:

jrainville: Oops. I commited everything. That's only needed for me, I'll remove

PR #449, opened 2019-10-07
Enable ethereum only when going to a page that needs it

jrainville:
Change `dappAutoEnable` to `false` and then only enable when needed.

I tried to not have to enable it in the offer list, but somehow it's needed, otherwise, I couldn't even get the offer list


jrainville:
Aah, yes that might make sense. I know I've had cases where just doing a `call`, I had to use a bogus accounts (randomly generated account with no ETH), but maybe `0x000` works too.
Let's indeed try

PR #450, opened 2019-10-08
feat: add country flags

jrainville:
image

### DIFF
--- a/src/js/services/googleMap.js
+++ b/src/js/services/googleMap.js
@@ -15,7 +15,7 @@ export function getLocation(place) {
       if (!res.length) {
         return reject(new Error(`No result for location ${place}`));
       }
-      resolve(res[0].location);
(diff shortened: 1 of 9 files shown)


src/js/services/googleMap.js
```diff
@@ -15,7 +15,7 @@ export function getLocation(place) {
       if (!res.length) {
         return reject(new Error(`No result for location ${place}`));
       }
-      resolve(res[0].location);
+      resolve({location: res[0].location, countryCode: res[0].country.short_name});
```
richard-ramos: Dumb question,  what happens if a user puts a location that doesn't resolve to a country code? would the offer still appear in the offer list?

jrainville: I'm fairly sure the offer still appears, but the location is bogus and there would be no country flag. I think Ricardo did that once (he put emojis I think haha) and we still had the offer in the list

PR #454, opened 2019-10-21
feat(dapp): remove seller license verification

richard-ramos:


richard-ramos:
Ah right! I did not have the problem with the wallet because I was using `master` and it doesnt have the EIP1102 change. Good catch!

jrainville:
Oh, my PR is not merged yet, makes sense. Should I merge it then you update, or you merge yours then I update?
I already have to change where I call the enable on the offerList, so it's not a problem for me

jrainville:
I'll let you merge yours first. I'll keep working on my old branch before merging it

richard-ramos:
I made some quick changes to remove some calls related to the license feature. However I think I'll not remove functionality from the license feature in case we decide to use Licenses for something.

PR #460, opened 2019-10-24
feat: slippage rate protection

richard-ramos:
- Adds a max slippage rate value
- Allows users to swap specific amounts


richard-ramos:
Slippage is what happens when a trade is done with a price different than the requested price.
I'm still not sure if it's necessary to add this protection here, because in practice any user could keep calling the swap function with smaller amounts and ignore the max slippage rate.

Maybe having a time delay to execute `swap` or limit the quantity of calls per day would be helpful.... I'm still researching! :)

PR #466, opened 2019-10-31
fix: Istanbul's EIP1884 may break contracts that receive eth in the fallback function

richard-ramos:


3esmit:
Seems good to me. 
Good to mention that solidity needs a better way of handling these now, because `destination.call.value(amount)("")` is awful, this probably will come soon and this could be improved.

bgits:
Not too familiar with the expected functionality of the different features here but for EIP1884 purposes LGTM.

PR #468, opened 2019-11-01
feat: add token icons in bg of landing

jrainville:
I forgot to put the little background icons
Also, some deps were missing because my yarn install was bugged

image


src/js/pages/Home/index.scss
```diff
@@ -21,8 +22,59 @@ $landing-text: #5C667D;
...
+    position: relative;
+    left: -231px;
+    margin-top: 60px;
+
+    @media screen and (max-width: 768px) {
+      width: auto;
+      left: 0;
+    }
+
+    &:before {
```
richard-ramos: Very clever!

jrainville: Thanks!
Not gonna lie, that was my girlfriend's idea hahaha. I asked her "what's the best way to have two background images" and she said "well, It's obvious, use before and after". I was like "yes, yes, very obvious :sweat_smile:"

PR #470, opened 2019-11-01
feat: create escrows without requiring signatures

richard-ramos:


richard-ramos:
Yeah, I'll probably ask for a proper review once I rebase this with all the changes from master.
With regards to the signature not being required. Originally I had that to avoid coupling the metadata and escrow contract, however working on the stake functionality made me realize that it wasn't necessary, because in this case the Buyer is the one starting the whole escrow process, so we might just take the `msg.sender` directly instead of obtaining it from the signature.

The signature functionality will still remain in the contract in case in the future we add some feature for the seller to start the escrow with the buyer's signature. Probably that will depend on a better integration with status once we have the messaging functionality available.

PR #472, opened 2019-11-06
fix: load user info on creating new escrow

richard-ramos:


richard-ramos:
I noticed that `state.metadata.users` did not have information for the buyer after creating a new escrow.

PR #488, opened 2019-11-14
fix: fix data script missing the stake amount for addOffer

jrainville:
Also only run script once


jrainville:
Added a second commit which adds back the arbitrations in the data script. I also fixed the nonce issues.
I found doing it that sometimes, the script hangs, but it's because of WS + Web3. Using Ganache, it doesn't hang and I couldn't test with RPC because I have a "too big" error.

PR #497, opened 2019-11-21
feat: ui changes to support different contact methods in profile edition and offer creation

richard-ramos:


src/js/utils/strings.js
```diff
@@ -4,3 +4,10 @@ export const formatArbitratorName = (user, address, noUsernameLabel, index) => {
   if (!user) return `${index >= 0 ? index + 1 + ' - ' : ""}` + address + ' - Loading...';
   return `${index >= 0 ? index + 1 + ' - ' : ""}${user.username || (noUsernameLabel ? noUsernameLabel : "No username available")} (${compactAddress(address, 3)})  ${user.location ? ' from ' + user.location : ''} - ${user.upCount || 0}↑  ${user.downCount || 0}↓`;
 };
+
+export const getContactDataItem = (contactData, item) => {
+  if(!contactData) return '';
+  contactData = contactData.split(':');
+  if(contactData[item]) return contactData[item];
```
jrainville: ~~I think we should use a word instead of item here or in the places where we call the function, use a constant, because `0` and `1` is not super intuitive.
So either:
getContactDataItem(contactData, 'method')
or
getContactDataItem(contactData, METHOD); // Where METHOD === 0~~

NVM, I just had another idea. The Function should instead return an object where there is `method` and `data` or whatever the name of the second part.

src/js/utils/strings.js
```diff
@@ -4,3 +4,10 @@ export const formatArbitratorName = (user, address, noUsernameLabel, index) => {
   if (!user) return `${index >= 0 ? index + 1 + ' - ' : ""}` + address + ' - Loading...';
   return `${index >= 0 ? index + 1 + ' - ' : ""}${user.username || (noUsernameLabel ? noUsernameLabel : "No username available")} (${compactAddress(address, 3)})  ${user.location ? ' from ' + user.location : ''} - ${user.upCount || 0}↑  ${user.downCount || 0}↓`;
 };
+
+export const getContactDataItem = (contactData, item) => {
```
jrainville: This makes me think, we should probably make a reverse function too where it generates the contact data string with the `:`, that way, the components don't have to know about it, they just give a method and a string.

PR #499, opened 2019-11-21
Add norification form in the Profile settings

jrainville:
Adds the Notification form in the Profile settings.
Adds the actions and sagas to subscribe and unsubscribe
Adds a page to verify the subscription
Later PRs will add this feature to the wizards, etc.


src/js/features/emailNotifications/saga.js
```diff
@@ -0,0 +1,89 @@
...
+import {call, fork, put, takeEvery} from 'redux-saga/effects';
+import {
+  CHECK_EMAIL_SUBSCRIPTION, CHECK_EMAIL_SUBSCRIPTION_SUCCESS, CHECK_EMAIL_SUBSCRIPTION_FAILURE,
+  SUBSCRIBE_EMAIL, SUBSCRIBE_EMAIL_FAILURE, SUBSCRIBE_EMAIL_SUCCESS,
+  VERIFY_EMAIL, VERIFY_EMAIL_FAILURE, VERIFY_EMAIL_SUCCESS,
+  UNSUBSCRIBE_EMAIL, UNSUBSCRIBE_EMAIL_FAILURE, UNSUBSCRIBE_EMAIL_SUCCESS
+} from './constants';
+import axios from 'axios';
+
+const DAPP_ID = 'teller-network';
```
richard-ramos: What do you think of moving the constants related to the service URL to a constants file inside `src/constants`?

src/js/pages/NotificationSettings/index.jsx
```diff
@@ -0,0 +1,131 @@
...
+      emailValid: false
+    };
+    this.props.checkSubscription();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.isSubscribed !== this.props.isSubscribed && this.props.isSubscribed === true) {
+      this.setState({showEmailSection: this.props.isSubscribed});
+    }
+    // TODO find a way to retrieve the email
```
richard-ramos: This one might be tricky. We need a way to safely return the email from the service to avoid anyone being able to know the user's email address by requesting it to the API using only the address.

Maybe the user needs to sign a phrase, send it to the API and if the signature is valid the email is returned?

Another possibility is displaying in the notification settings that the user is subscribed, but not showing the email.

jrainville: Like you said, we could just not show the email field, it's really not the end of the world, it's a nice to have.
I guess we could do something similar to the unsubscribe, where we send the Dapp ID signed, with the account, so the API can know that it's us, but requiring the user to sign something just for a "get" sounds really bad.
Let's just leave it as a TODO for now

PR #504, opened 2019-11-22
Add notification form for the seller and in escrow

jrainville:
Adds a notification for page that you can acces at the end of the create offer wizard and in the escrow page


embarkConfig/data.js
```diff
@@ -165,8 +165,7 @@ module.exports = async (licensePrice, arbitrationLicensePrice, feeMilliPercent,
         tokens[1],
         // TODO un hardcode token and add `approve` in the escrow creation below
         // tokens[Math.floor(Math.random() * tokens.length)],
-        address,
-        address,
+        'Status:' + address,
```
richard-ramos: We can change this to `getContactData('Status', address)`

PR #505, opened 2019-11-22
feat: contact details modal dialog when clicking "Chat with buyer/seller"

richard-ramos:


src/css/bootstrap-overrides.scss
```diff
@@ -181,7 +181,6 @@ $input-icon-height: 20px;
 
 .modal {
   .modal-dialog {
-    margin: 80px 32px 0;
```
jrainville: I fixed that problem in my branch. Just replace that by `margin-top: 80px` or don't touch it at all

src/js/pages/Escrow/index.jsx
```diff
@@ -37,9 +37,18 @@ import events from '../../features/events';
...
 
 const {toBN} = web3.utils;
 
+const ContactMethods = {
+  'Status': 'Status chat key',
+  'Telegram': 'Telegram ID',
+  'Line': 'Line ID',
+  'KakaoTalk': 'KakaoTalk ID',
+  'WeChat': 'WeChat ID'
+};
```
jrainville: We should put those in one of the constants file

PR #511, opened 2019-11-26
Add dispute motives when opening a case

richard-ramos:
screencapture-localhost-3000-2019-11-26-15_19_14


src/js/pages/Arbitration/index.jsx
```diff
@@ -225,3 +225,20 @@ export default connect(
     getProfile: metadata.actions.loadUserOnly
   }
 )(withRouter(Arbitration));
+
+
+function getMotiveTxt(escrow, openedByBuyer) {
+  switch (escrow.arbitration.motive) {
+    case UNRESPONSIVE:
+      if (openedByBuyer) return 'Seller is unresponsive/inactive after payment is made';
+      return 'Buyer is unresponsive/inactive after marking the transaction as paid';
```
jrainville: Can make this sleek by doing something like:
```
return `${openedByBuyer ? 'Buyer' : 'Seller'} is unresponsive/inactive after marking the transaction as paid`;
```

src/js/pages/OpenDispute/index.jsx
```diff
@@ -67,30 +75,34 @@ class OpenDispute extends Component {
...
+          { !isBuyer && <span className="text-muted text-small"><br />buyer has marked trade as paid but is unresponsive and inactive</span>}
+          </CheckButton>
+          <CheckButton size="l" active={this.state.motive === PAYMENT} onClick={this.setMotive(PAYMENT)}>
+          <b>Payment issue</b>
+          { isBuyer && <span className="text-muted text-small"><br />you have made the payment, seller is responsive but states that something is wrong with the payment process, refuses to release</span> }
+          { !isBuyer && <span className="text-muted text-small"><br />buyer is active, has made an attempt to pay, but there are issues with the payment</span>}
+          </CheckButton>
+          <CheckButton size="l" active={this.state.motive === OTHER} onClick={this.setMotive(OTHER)}>
+          <b>Other</b>
+          { isBuyer && <span className="text-muted text-small"><br />any other reason</span> }
```
jrainville: Comment for this whole section: It's better to use `<p>`s instead or `<br/>`s, because they are easier to style and just cleaner.
Anyway, I see you already wrapped your text in `<span>`s, so just need to replace by `<p>`
If you don't want the huge margin under the `<p>`s, you can just put the `mb-0` classname.
Same thing for the bold tags, it's better to do that with CSS, or in our case, with a `<span className="font-weight-bold">`, what way we can control the size.

Last comment, it's prettier to start the sentences with Capital letters

PR #521, opened 2019-12-04
fix: hide filters on 'My trades'

richard-ramos:


src/js/pages/MyTrades/components/Trades.jsx
```diff
@@ -113,7 +112,7 @@ class Trades extends Component {
           </FormGroup>
         </Form>}
 
-        {!this.state.showFilters && <p className="text-small text-muted mb-1 clickable" onClick={() => this.setState({showFilters: true})}>Show Filters <FontAwesomeIcon icon={faCaretDown}/></p>}
+        {false && <p className="text-small text-muted mb-1 clickable" onClick={() => this.setState({showFilters: true})}>Show Filters <FontAwesomeIcon icon={faCaretDown}/></p>}
```
jrainville: Might as well delete. We have git to add it back if we ever need

src/js/pages/MyTrades/components/Trades.jsx
```diff
@@ -125,7 +124,7 @@ class Trades extends Component {
               const isBuyer = addressCompare(trade.buyer, this.props.address);
               const tradeStyle = getTradeStyle(trade, isBuyer);
 
-              return (<Card key={index} className="mb-3 shadow border-0 offer-card"
+              return (<Card key={index} className={classnames("mb-3", "shadow", "border-0", "offer-card", {"card-transparent": !this.props.active})}
```
jrainville: Just a tip. You can do:
```suggestion
              return (<Card key={index} className={classnames("mb-3 shadow border-0 offer-card", {"card-transparent": !this.props.active})}
```
Up to you

PR #549, opened 2019-12-09
fix: location and amount filter refs

richard-ramos:


src/js/pages/OffersList/components/SorterFilter.jsx
```diff
@@ -165,10 +165,12 @@ class AmountModal extends Component {
...
       amount: props.amount === -1 ? '' : props.amount
     };
+
+    this.amountInput = React.createRef();
   }
 
-  componentDidMount() {
-    this.amountInput.focus();
+  componentDidUpdate() {
+    this.amountInput.current.focus();
```
jrainville: Doing this on every update seems intense. Won't it refocus all the time?

richard-ramos: hmm... well as far as my extensive tests went (of opening the dialog and writing values 2 or 3 times) I saw no issue. I tried using componentDidMount, and the input variables were always null.

Not sure how to fix this without using `componentDidUpdate`

jrainville: Alright, since it's a modal, I guess it's fine

PR #555, opened 2019-12-10
fix: subscribing to emails gets stuck

richard-ramos:


src/js/features/emailNotifications/reducer.js
```diff
@@ -41,7 +41,6 @@ function reducer(state = DEFAULT_STATE, action) {
     case SUBSCRIBE_EMAIL_SUCCESS:
       return {
         ...state, ...{
-          isSubscribed: true,
```
jrainville: Why remove that?

richard-ramos: Because something weird happens if I subscribe to notifications, go to "My profile" and then I go back to "Notification settings".
```js
uncaught at checkSubscription TypeError: Cannot convert undefined or null to object
    at Function.from (<anonymous>)
    at i (webpack-internal:///./node_modules/react-validation/build/form.js:67:20)
...
index.js:1437 The above error occurred in the <Control(s)> component:
    in Control(s) (at NotificationForm/index.jsx:29)
    in div (created by FormGroup)
    in FormGroup (at NotificationForm/index.jsx:27)
    in form (created by t)
    in t (created by Form(t))
```

jrainville: Hmm, that's weird. Oh well.

PR #556, opened 2019-12-10
Add location settings pages

jrainville:
Edit: there is a commit for the new margin selector in this because I'm a dum dum
image
image


jrainville:
Welp, I messed up and pushed my new commit for the margin in this PR.
I'm too lazy to fix it, so please review it all together :)

contracts/teller-network/MetadataStore.sol
```diff
@@ -338,7 +336,7 @@ contract MetadataStore is Stakable, MessageSigned, SecuredFunctions {
     function offer(uint256 _id) external view returns (
         address asset,
         string memory currency,
-        int8 margin,
+        int16 margin,
```
richard-ramos: With this change we're allowing a margin % that goes from -32,768 to +32,767.  Do you think w should limit the max and min value in a condition in the offer creation function?

src/js/wizards/Sell/6_Margin/components/MarginSelectorForm.jsx
```diff
@@ -46,21 +53,19 @@ class MarginSelectorForm extends Component {
         <FormGroup className="mb-0">
           <Row>
             <Col md={9} sm={9} xs={8}>
-              <Slider className="mb-3 p-4" min={-99} max={100} defaultValue={0}
-                      onChange={(value) => this.onMarginChange(value)} value={margin}/>
+              <Slider className="mb-3 p-4" min={-100} max={100} defaultValue={1}
```
richard-ramos: -100% in this case would mean giving away the crypto for free :)

contracts/teller-network/MetadataStore.sol
jrainville: You'd say it's better to do it in the contract, in the form or both?

src/js/wizards/Sell/6_Margin/components/MarginSelectorForm.jsx
jrainville: Oops 🤦‍♂

contracts/teller-network/MetadataStore.sol
richard-ramos: Hmm... in the smart contract it would be a nice to have but it's not super important, because it's only an informative value. I wonder what would happen if someone puts -101% and a price of 50usd for some crypto? Would the UI display a minimum limit of -0.5usd?

And in the case of the form, yeah, I think we should have some valid limits, starting from -99 to some upper limit %?

PR #560, opened 2019-12-13
feat: Allow specifying different destination for escrow funds

richard-ramos:
UI will keep the same behavior as before


richard-ramos:
I did not try TBH because I was worried about the increase in size of the contract due to having extra functions to handle the optional parameter.

During the day I'll try that and compare both solutions.

PR #582, opened 2019-12-27
A batch of UI fixes

jrainville:
See commit messages to see what was fixed


src/js/pages/Escrow/index.jsx
```diff
@@ -21,7 +21,7 @@ import ErrorInformation from "../../components/ErrorInformation";
 import {Col, Row, Input, Button} from 'reactstrap';
 import RoundedIcon from "../../ui/RoundedIcon";
 import exclamationCircle from "../../../images/exclamation-circle.png";
-import checkCircle from "../../../images/check-circle.svg";
+import {ReactComponent as checkCircle} from '../../../images/check-circle.svg';
```
richard-ramos: How does this work?

jrainville: This is the way to get the actual SVG data from the image.
i.e. when you do `import exclamationCircle from "../../../images/exclamation-circle.png";`, it only gets the path to the image, so that you can put the path in the `src` attribute of the ``.
However, this means that you cannot use the `fill` css attribute to change the color of the image.

When using `ReactComponent`, it renders as the `<svg>` tag, on which  you can change the `fill` attribute (and others if needed.

TLDR: it's weirder, but gives you more flexibility on the image (when SVG at least)

PR #589, opened 2020-01-06
fix: format fiat prices for offers and trades

jrainville:


src/js/utils/numbers.js
```diff
@@ -18,6 +18,13 @@ export function limitDecimals(input, numDecimals = 6) {
   return parseFloat(input.toFixed(numDecimals));
 }
 
+export function formatFiatPrice(price) {
+  if (price < 0.01) {
+    return '< 0.01';
```
richard-ramos: :thinking:

jrainville: Any problem with it or you're just meme-ing? :P

richard-ramos: Wouldnt this always show < 0.01 for assets like SNT?

richard-ramos: I found Simon's suggestion on how to handle these small numbers:
```
0.008200 display as 0.008
0.000100 display as 0.0001
0.129019 display as 0.12
0.000123 display as 0.0001
```

jrainville: If the price of SNT goes higher than 0.01, no, but yes, it will often show that

On the trade page (the next page when clicking the offer), it shows with 4 digits, so you can know the real price.

Might be worth checking with Simon to see if in those cases we should show more digits?

jrainville: Oh, you answered while I was writing.
I'll update the PR then

PR #597, opened 2020-01-07
fix: stake price in offer summary and ETH name

richard-ramos:


src/js/utils/networks.js
```diff
@@ -615,7 +615,7 @@ export const Tokens = {
   'rinkeby': [
     {
       symbol: 'ETH',
-      name: 'ETH',
+      name: 'Ethereum',
```
jrainville: Not sure, but this might cause Coin Gecko issues.
Long story short, I use the symbol and name to check the coin gecko ID and that's why I changed some of the names to the Gecko ones. That's because I use a binary search instead of a full search (O(log(n)) instead of O(n))

Anyway, to test, just reset your cache and check if the price for ETH works (check console too)

src/js/wizards/Sell/8_Summary/components/FinalModal.jsx
```diff
@@ -9,14 +9,18 @@ import {getTokenImage} from "../../../../utils/images";
...
+
+  return (
   <Modal isOpen={isOpen} toggle={hide} className="final-step-modal">
     <ModalHeader>
       <RoundedIcon image={pencilIcon} className="mb-3" bgColor="blue"/>
       <p className="m-0">Final Step</p>
     </ModalHeader>
     <ModalBody className="text-center">
-      <p className="text-muted">Almost there! As a way to protect teller from spams, each offer is under a stake condition. The stake is as small as 0.10 USD in ETH.</p>
+      <p className="text-muted">Almost there! As a way to protect teller from spams, each offer is under a stake condition. The stake is as small as 0.01 ETH.</p>
```
jrainville: Should add this to the i18n file

src/js/utils/networks.js
richard-ramos: Looks like it works with 'Ethereum'

jrainville: Good

PR #599, opened 2020-01-08
fix: calculate function signatures automatically

richard-ramos:


src/js/provider.js
```diff
@@ -7,17 +8,20 @@ import SNT from '../embarkArtifacts/contracts/SNT';
 import {checkNotEnoughETH} from './utils/transaction';
 import {addressCompare, zeroAddress} from './utils/address';
 import {canRelay} from './features/escrow/helpers';
+import stripHexPrefix from 'strip-hex-prefix';
```
jrainville: Just to be sure, this was already installed or you forgot to add it to package.json?

richard-ramos: I saw it installed as a transitive dependency in `yarn.lock`

jrainville: If it's just a transitive dep, we should add it to package.json in case the dependency we have that uses it removes it or changes the version to something that breaks.

PR #615, opened 2020-01-13
feat: new mainnet deployment

richard-ramos:


richard-ramos:
@jrainville, I couldn't get the ENSRegistry to appear in shared.mainnet.json, so the deployment will not resolve ENS. Let's check this tomorrow to redeploy the site.

richard-ramos:
I did not see the ENSRegistry file in embarkArtifacts, and something curious is that I checked in the old mainnet chains file and it has ENSRegistry in its list: https://github.com/status-im/status-teller-network/pull/615/files#diff-84089144bc20e8ea66ca3d3955fc23fdR5

Looks like I have a bug :D  
I think I can 'fix' the deployment by copying the ENSRegisry files from the previous deployment and adding it too to the chain file.

jrainville:
That's very weird, you didn't change the Embark version I think?
Maybe something changed in the namesystem config?

PR #626, opened 2020-01-15
fix problems when prices don't load or load partially

jrainville:
This started as trying to fix the Loading Data button never changing and evolved into an effort to fix when prices don't load to fix some of the crashes we have

### DIFF
--- a/src/js/wizards/Sell/2_Currency/components/FiatSelectorForm.jsx
+++ b/src/js/wizards/Sell/2_Currency/components/FiatSelectorForm.jsx
@@ -28,22 +28,19 @@ class FiatSelectorForm extends Component {
     }
 
     return (
-      <Fragment>
-        <h2>{t('sellerFiatContainer.title')}</h2>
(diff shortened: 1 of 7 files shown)


src/js/wizards/Sell/2_Currency/components/FiatSelectorForm.jsx
```diff
@@ -28,22 +28,19 @@ class FiatSelectorForm extends Component {
...
-            placeholder={t("fiatSelectorForm.placeholder")}
-            onInputChange={this.onInputChange}
-            submitFormOnEnter={true}
-            emptyLabel={t("fiatSelectorForm.emptyLabel")}
-            defaultSelected={defaultSelectedValue}
-          />
-        </FormGroup>
-      </Fragment>
+      <FormGroup>
+        <Label className="text-small mt-3 mb-0">Local currency</Label>
```
richard-ramos: i18n

PR #638, opened 2020-01-17
feat: add row in escrow to approve tokens

jrainville:
This step is only for the seller. The buyer will only see the funding step

image


hesterbruikman:
@Jo#6415 from a conversation with @3esmit  just now on a UI bug in Assemble: Can the extra step in 'add row to approve tokens' be dependent on currency? As ETH doesn't require the authorization step tokens do. 
cc @simonricoo

jrainville:
@hesterbruikman as stated in the call, the code makes sure to add the extra row only for ERC20 escrows (aka not for ETH escrows)

PR #653, opened 2020-01-22
feat(contracts): limit offers per seller

richard-ramos:


contracts/teller-network/OfferStore.sol
```diff
@@ -84,6 +87,7 @@ contract OfferStore is Stakable, SecuredFunctions, Proxiable {
         burnAddress = _burnAddress;
 
         basePrice = 0.01 ether;
+        maxOffers = 10;
```
jrainville: Why set it to 0 again here? Since it is already initialized at 10

contracts/teller-network/OfferStore.sol
```diff
@@ -164,6 +181,7 @@ contract OfferStore is Stakable, SecuredFunctions, Proxiable {
         uint256 offerId = offers.push(newOffer) - 1;
         offerWhitelist[msg.sender][offerId] = true;
         addressToOffers[msg.sender].push(offerId);
+        offerCnt[msg.sender]++;
```
jrainville: I guess when `offerCnt[msg.sender]` is undefined, it starts at 0?

richard-ramos: yes

contracts/teller-network/OfferStore.sol
richard-ramos: If you use the offerstore contract directly, maxOffers will be initialized by 10 
```
    uint public maxOffers = 10;
```

However in the case of proxyes, that initialization does not work because it will not be the OfferStore contract the one keeping the state, but the proxy, hence the need of an init() function that initializes any state variable.

jrainville: Thanks TIL

PR #657, opened 2020-01-23
feat: USD prices for stakes

richard-ramos:


contracts/common/USDStakable.sol
```diff
@@ -0,0 +1,49 @@
...
+import "./Medianizer.sol";
+
+
+contract USDStakable is Stakable {
+    Medianizer public medianizer;
+    uint public basePrice;
+
+    constructor(address payable _burnAddress, address _medianizer) public Stakable(_burnAddress) {
+        burnAddress = _burnAddress;
+        basePrice = 1 ether;  // 1 usd
```
jrainville: I think you messed up here

jrainville: Ah ok, nvm I understand

PR #681, opened 2020-01-28
feat: fallback arbitrator UI

richard-ramos:
i18n was not applied due to this page not being the final design. Possibly we'll require integration with Gnosis Multisig


src/js/features/arbitration/actions.js
```diff
@@ -29,7 +29,7 @@ import EscrowProxy from '../../../embarkArtifacts/contracts/EscrowProxy';
 ArbitrationLicense.options.address = ArbitrationLicenseProxy.options.address;
 Escrow.options.address = EscrowProxy.options.address;
 
-export const getDisputedEscrows = () => ({type: GET_DISPUTED_ESCROWS});
+export const getDisputedEscrows = (fallbackDisputes = false) => ({type: GET_DISPUTED_ESCROWS, fallbackDisputes});
```
jrainville: The flag name is not very clear. It took me a while to get what it meant. It should probably be `inludeFallbackDisputes`

PR #691, opened 2020-02-05
feat: update escrow destination address

richard-ramos:


contracts/teller-network/Escrow.sol
```diff
@@ -284,6 +284,20 @@ contract Escrow is IEscrow, Pausable, MessageSigned, Fees, Arbitrable, Proxiable
...
     }
 
+    function updateDestination(uint _escrowId, address payable _destination) external whenNotPaused {
+        EscrowTransaction storage trx = transactions[_escrowId];
+
+        EscrowStatus mStatus = trx.status;
+        require(mStatus == EscrowStatus.FUNDED || mStatus == EscrowStatus.CREATED || mStatus == EscrowStatus.PAID,
+                "Only transactions in created, funded or paid state can have their destination updated");
+
+        require(trx.buyer == msg.sender || trx.destination == msg.sender, "Only the buyer or the destination address can invoke this function");
```
jrainville: I really don't think the destination should invoke that. Otherwise, I could call this function on random escrows and get free money

richard-ramos: Yeah, i think I will remove that because maybe it does not make sense for the receiver address to be able to change itself so the escrow funds are sent somewhere else, but... wouldn't `trx.destination == msg.sender` limit the function to be called only by the destination address that was set on that particular escrow?

jrainville: Not sure what you mean, but what I meant was that only the OG buyer should be able to invoke this function, i.e. only keep `trx.buyer == msg.sender`.

But now I just realized I'm a bit dumb. I thought the condition read "if you are the account of the NEW destination", but it's "if you are the account of the OLD destination", so I guess it's fine either way. There will probably few cases where the buyer != the destination and even less where the destination wants to switch it to someone else without being the buyer.

TLDR: I'm dumb, and it's fine as is or you can change it, whatever

PR #699, opened 2020-02-11
feat: show fallback arbitration options only to specific accounts in the multisig

richard-ramos:


embarkConfig/contracts.js
```diff
@@ -256,6 +262,10 @@ module.exports = {
       },
       KyberFeeBurner: {
         address: "0x7702CaaE3D8feE750c4464d80FCb14Ce05e00743"
+      },
+      GnosisSafe: {
+        address: FALLBACK_ARBITRATOR_MAINNET,
+        abiDefinition: [{"constant":true,"inputs":[{"name":"owner","type":"address"}],"name":"isOwner","outputs":[{"name":"","type":"bool"}],"payable":false,"stateMutability":"view","type":"function"}]
```
jrainville: We should put this in  a constant and reuse it at both spots

PR #711, opened 2020-02-28
feat: show a warning on signatures and add a doSign function

jrainville:
Similar to the check before sending a TX, I added a `doSign`, that shows a warning about Signing and the  signs.


src/js/components/TransactionWarning/index.jsx
```diff
@@ -8,15 +8,33 @@ import ModalDialog from "../ModalDialog";
...
+    case GSN:
+      message = t('transactionWarning.walletWillAskGSN');
+      break;
+    case SIGN:
+      message = t('transactionWarning.walletSignature');
+      break;
+    default:
+      message = t('transactionWarning.walletWillAsk');
+  }
+  console.log('TYPE???', warningType);
```
richard-ramos: :scissors:

PR #718, opened 2020-06-02
fix use of legacy get.status.im domain

jakubgs:
We've replaced https://get.status.im/ with https://join.status.im/ for deep links.

### DIFF
--- a/SWARM.md
+++ b/SWARM.md
@@ -18,7 +18,7 @@ contributors:
 
 ## Communication
 (required)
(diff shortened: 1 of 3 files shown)


jakubgs:
@jrainville should I care that all the CI is red?

jrainville:
@jakubgs ah no don't worry. I think one of the contract test fails because of a small mistake. @richard-ramos will fix it soon. Anyway, it's not like your change has an impact on our CI.

PR #720, opened 2020-06-04
Remove requires for safeTransfer

richard-ramos:


richard-ramos:
Openzepellin's SafeTransfer do not return a boolean value, and already have requires that check if a transfer was successful:
https://github.com/status-im/status-teller-network/blob/fix/safe_transfer/contracts/token/SafeTransfer.sol#L35-L43

jrainville:
Cool, thanks for the explanation

